PR #4, opened 2024-09-27
Support RPC in service bindings

### LINKED ISSUE #1: Support worker entries that extend `WorkerEntrypoint`

### LINKED ISSUE #2: Support RPC via Service Bindings

jamesopstad:
resolves #1, resolves #2

This PR adds support for RPC via service bindings and workers defined as classes. It does this by using a class that extends `WorkerEntrypoint` as the wrapping worker that contains the module runner. The class constructor returns a `Proxy` that wraps the class and intercepts any method calls. These are then forwarded to the user code.

### DIFF
--- a/packages/vite-plugin-cloudflare/src/runner/worker.ts
+++ b/packages/vite-plugin-cloudflare/src/runner/worker.ts
@@ -1,3 +1,4 @@
+import { WorkerEntrypoint } from 'cloudflare:workers';
 import { ModuleRunner } from 'vite/module-runner';
 import { UNKNOWN_HOST, INIT_PATH } from '../shared.js';
 import type { FetchResult } from 'vite/module-runner';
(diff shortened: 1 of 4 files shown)


packages/vite-plugin-cloudflare/src/runner/worker.ts
```diff
@@ -69,50 +70,120 @@ function createModuleRunner(env: RunnerEnv, webSocket: WebSocket) {
...
+
+						if (url.pathname === INIT_PATH) {
+							return target.#init.apply(target, [request]);
+						}
+
+						return target.#fetch.apply(target, [request]);
+					};
+				} else {
+					return (...args: unknown[]) => {
+						return target.#rpc.apply(target, [prop, args]);
```
dario-piotrowicz, line 91: this assumes that all the properties of a `WorkerEntrypoint` have to be methods, is that tha case? can't I simply expose a constant from my class instance?

dario-piotrowicz (approved):
The solution makes sense to me 🙂 

as I mentioned the only thing I am not sure about is the assumption that all fields of a worker exported instance need to be methods (which I am not sure if it's correct)

packages/vite-plugin-cloudflare/src/runner/worker.ts
jamesopstad: Good point. I thought you could only call methods over RPC but it turns out you can also access getters - https://developers.cloudflare.com/workers/runtime-apis/rpc/visibility/#class-instance-properties. I'll add support for this.

jamesopstad:
Going to merge this for now and add a separate issue for supporting getters.

PR #15, opened 2024-10-08
integrate module resolution in `vite-plugin-cloudflare`

dario-piotrowicz:
integrate the module resolution solution implemented in `@flarelabs-net/vite-environment-provider-cloudflare` in the new `vite-plugin-cloudflare` package

also add the module-resolution tests that were present in the old repository here too


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -100,6 +135,11 @@ export function cloudflare<
 
 					return {
 						...workerOptions,
+						compatibilityFlags: [
+							...(workerOptions.compatibilityFlags ?? []),
+							'nodejs_compat',
```
petebacondarwin, line 140: OOC - why do we need to add this flag?

petebacondarwin:
```suggestion
							...(workerOptions.compatibilityFlags?.includes('nodejs_compat')
								? ['nodejs_compat']
								: []),
```

dario-piotrowicz: oops, no sorry I added this by mistake, thanks for spotting this! 🙇

dario-piotrowicz: (by the way, it was when I was investigating this: https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/15/files#diff-544c6cfcc950af42d4e59824927b6ba7dd94705534aa7df3ea200f339b3c9d3eR8-R15)

dario-piotrowicz: removed 👍

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -110,37 +146,57 @@ export function cloudflare<
...
+										string,
+										string,
+										FetchFunctionOptions,
+									];
+
+								if (moduleId.startsWith('cloudflare:')) {
+									const result = {
+										externalize: moduleId,
+										type: 'builtin',
+									} satisfies vite.FetchResult;
```
petebacondarwin, line 172: Can this externalization not be handled by configuring Vite rather than hard coding here?

packages/vite-plugin-cloudflare/src/module-fallback.ts
```diff
@@ -0,0 +1,385 @@
...
+
+export type ResolveIdFunction = (
+	id: string,
+	importer?: string,
+	options?: {
+		resolveMethod: 'require' | 'import';
+	},
+) => Promise<string | undefined>;
+
+export function getModuleFallbackCallback(
```
petebacondarwin, line 17: NIT: the type of the return value of this function is `ModuleCallbackHandler`, so I expected this method to be called `getModuleFallbackHandler()`.

But perhaps you just liked the poetic nature of fallback-callback? 😄

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -11,6 +11,11 @@ import type {
 	CloudflareEnvironmentOptions,
 	CloudflareDevEnvironment,
 } from './cloudflare-environment';
+import {
+	getModuleFallbackCallback,
+	type ResolveIdFunction,
```
jamesopstad, line 16: Is it OK if we keep type imports separate and below the other imports? I just find it easier to scan.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -11,6 +11,11 @@ import type {
...
+} from './module-fallback';
+import { createIdResolver } from 'vite';
```
jamesopstad, line 18: Vite is already imported so this can be used as `vite.createIdResolver`.

```suggestion
```

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -83,6 +88,36 @@ export function cloudflare<
 				}
 			}
 
+			const esmResolveId = createIdResolver(viteConfig, {});
```
jamesopstad, line 91:
```suggestion
			const esmResolveId = vite.createIdResolver(viteConfig, {});
```

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -83,6 +88,36 @@ export function cloudflare<
...
+
+			// for `require` calls we want a resolver that prioritized node/cjs modules
+			const cjsResolveId = createIdResolver(viteConfig, {
```
jamesopstad, line 94:
```suggestion
			const cjsResolveId = vite.createIdResolver(viteConfig, {
```

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -83,6 +88,36 @@ export function cloudflare<
...
+				id,
+				importer,
+				{ resolveMethod } = {
+					resolveMethod: 'import',
+				},
+			) => {
+				const resolveIdFn =
+					resolveMethod === 'import' ? esmResolveId : cjsResolveId;
+
+				const firstWorkerName = Object.keys(pluginConfig.workers)[0]!;
```
jamesopstad, line 112: Let's inlcude a comment here so we know this is temporary.

```suggestion
        // TODO: investigate making `resolveId` function unique for each worker
				const firstWorkerName = Object.keys(pluginConfig.workers)[0]!;
```

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -110,37 +146,57 @@ export function cloudflare<
 								type: 'ESModule',
 								path: runnerPath,
 							},
+							{
+								// we declare the workerd-custom-import as a CommonJS module, thanks to this
+								// require is made available in the module and we are able to handle cjs imports, etc...
+								type: 'CommonJS',
+								path: fileURLToPath(
+									new URL('workerd-custom-import.cjs', import.meta.url),
```
jamesopstad, line 154: Let's put this path in a variable at the top of the file with the others.

packages/vite-plugin-cloudflare/src/module-fallback.ts
```diff
@@ -0,0 +1,385 @@
+import { Request, Response } from 'miniflare';
+
+import { readFile, stat } from 'node:fs/promises';
+import { URL } from 'url';
```
jamesopstad, line 4: This import shouldn't be needed

packages/vite-plugin-cloudflare/src/runner/module-runner.ts
```diff
@@ -5,11 +5,24 @@ import type { WrapperEnv } from './env';
 
 let moduleRunner: ModuleRunner;
 
-export function createModuleRunner(env: WrapperEnv, webSocket: WebSocket) {
+// node modules using process.env don't find process in the global scope for some reason...
+// so let's define it here ¯\_(ツ)_/¯
+(globalThis as Record<string, unknown>).process = { env: {} };
```
jamesopstad, line 10: Not sure if this is the right thing to do here. Should we only provide `process.env` if `node_compat` is enabled or do you think it's still needed otherwise?

packages/vite-plugin-cloudflare/src/runner/module-runner.ts
```diff
@@ -5,11 +5,24 @@ import type { WrapperEnv } from './env';
...
+export async function createModuleRunner(
+	env: WrapperEnv,
+	webSocket: WebSocket,
+) {
 	if (moduleRunner) {
 		throw new Error('Runner already initialized');
 	}
 
+	// we store the custom import file path in a variable to skip esbuild's import resolution
+	const workerdReqImport = '../workerd-custom-import.cjs';
```
jamesopstad, line 21: Let's move this import inside the `runner` folder:

```suggestion
	const workerdReqImport = './workerd-custom-import.cjs';
```

packages/vite-plugin-cloudflare/src/workerd-custom-import.cts
```diff
@@ -0,0 +1,10 @@
+module.exports = async (path: string) => {
+  const result = await import(path);
+  // CJS modules don't provide named exports properly in workerd, the named values can be
+  // found under `default` instead, so here we spread default in the returned object
+  // (this should get fixed soon: https://github.com/cloudflare/workerd/pull/2194)
```
jamesopstad, line 5: Is this actually a workerd issue though? I thought named exports from CJS modules are always under `default` when imported in ESM.

packages/vite-plugin-cloudflare/tsup.config.ts
```diff
@@ -18,4 +18,11 @@ export default defineConfig([
 		noExternal: ['vite/module-runner'],
 		tsconfig: 'tsconfig.runner.json',
 	},
+	{
+		entry: ['src/workerd-custom-import.cts'],
+		outDir: 'dist',
```
jamesopstad, line 23: Let's move this inside the runner folder as that's where it is used:

```suggestion
		outDir: 'dist/runner',
```

tests/integration/src/fixtures/module-resolution/index.ts
```diff
@@ -0,0 +1,25 @@
...
+		const path = url.pathname;
+
+		if (allowedPaths.has(path)) {
+			const mod = await import(/* @vite-ignore */ `./src${path}`);
+			return Response.json(mod.default);
+		}
+
+		return new Response(`path not found: '${path}'`, { status: 404 });
+	},
+};
```
jamesopstad, line 25: Let's use the `ExportedHandler` type here:

```suggestion
export default {
	async fetch(request) {
		const url = new URL(request.url);
		const path = url.pathname;

		if (allowedPaths.has(path)) {
			const mod = await import(/* @vite-ignore */ `./src${path}`);
			return Response.json(mod.default);
		}

		return new Response(`path not found: '${path}'`, { status: 404 });
	},
} satisfies ExportedHandler;
```

jamesopstad (changes requested):
A few small suggestions but looks good. One general stylistic point is that I prefer to avoid explicit return types if possible. I feel that most of the time it's better to just let the type system do it's job. By providing return types you can also inadvertently widen the type.

packages/vite-plugin-cloudflare/src/module-fallback.ts
petebacondarwin: One more NIT: I feel like moving the actual fallback handler into a named function would make the code easier to follow:

```ts
export function getModuleFallbackCallback(
	resolveId: ResolveIdFunction,
): ModuleFallbackHandler {
	return patchModuleFallbackHandler(moduleFallbackHandler);

	async function moduleFallbackHandler(request: Request): Promise<Response> {
    ...
  }
}
```

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: sorry this got auto imported and I missed it 👍

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: mh.... that's a good point! 🤔

I think we can, I'll have a look 🙂👍

dario-piotrowicz:
> One general stylistic point is that I prefer to avoid explicit return types if possible. I feel that most of the time it's better to just let the type system do it's job. By providing return types you can also inadvertently widen the type.

this is exactly the opposite of my preference 😅

I think pretty much everyone agrees that relying on the type system automatically inferring types is the way to go, but when it comes to return types I feel like having the extra explicitly is actually quite good since there you can convey intent and let typescript (and other developers) know what you want your function to do (by having ts automatically infer the return type you lose a certain degree of control, and the code saved is negligibile IMO)

that's my personal view, but anyways I'm fine with going with whatever the team prefers 👍

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: We also need to add them (`cloudflare:workers`, `cloudflare:sockets` etc.) to `resolve.external` on the environment so that they will be externalised in the build but I'm not sure that will replace this behaviour in dev. Worth trying but if not let's keep build only stuff for a separate PR.

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: Sure, I don't have a particular preference

Sorry for not picking up the consistency you're introduced of using the top-level version 🙇 

PS: as a followup I can add this eslint rule: https://typescript-eslint.io/rules/consistent-type-imports/#fixstyle to make sure we keep all imports consistent in this regard

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: what about this? 🙂 
```suggestion
				// TODO: we only have a single module resolution strategy shared across all workers
				//       (generated using the first worker's dev environment)
				//       we should investigate and ideally have potential different resolutions per worker
				//       see: https://github.com/flarelabs-net/vite-plugin-cloudflare/issues/19
```

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: https://github.com/cloudflare/workers-sdk/blob/main/.prettierrc

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: done 👍

jamesopstad:
> > One general stylistic point is that I prefer to avoid explicit return types if possible. I feel that most of the time it's better to just let the type system do it's job. By providing return types you can also inadvertently widen the type.
> 
> this is exactly the opposite of my preference 😅
> 
> I think pretty much everyone agrees that relying on the type system automatically inferring types is the way to go, but when it comes to return types I feel like having the extra explicitly is actually quite good since there you can convey intent and let typescript (and other developers) know what you want your function to do (by having ts automatically infer the return type you lose a certain degree of control, and the code saved is negligibile IMO)
> 
> that's my personal view, but anyways I'm fine with going with whatever the team prefers 👍

Haha. Yeah, I had feeling we would disagree about this one. You may well be right and it's also probably a bit dependent on the context. Let's not worry about it in this PR anyway.

packages/vite-plugin-cloudflare/src/module-fallback.ts
dario-piotrowicz: oh! good spot! 😄 :+1:

packages/vite-plugin-cloudflare/src/runner/module-runner.ts
dario-piotrowicz: yeah, I'd personally just keep it here for now and then wait for @petebacondarwin's nodejs changes and potentially make this like obsolete from those and remove it then

But I am also totally fine removing this for now and skip the tests that fail because of the missing `process`

Just let me know your preference 🙂👍

jamesopstad: What do you think @petebacondarwin? Do you want to keep it for now and then replace it later?

packages/vite-plugin-cloudflare/tsup.config.ts
dario-piotrowicz: good point, done 👍

packages/vite-plugin-cloudflare/src/runner/module-runner.ts
dario-piotrowicz: good point, done 👍

vicb:
Sorry for a not really helpful comments but some comments on what I have read here:
- I do prefer explicit return type - that with a good function + param names allow to skip having to read the fn body
- +10 for a common eslint/prettier style that we would share across projects

I was thinking starting on a common style at the end of this week.

A few things I'd like to include is import sorting, force curly braces, ...

I have opinions about code style but very few are strong and that's something we should discuss. Generally I like to stick to something "common" and avoid exotic settings. Also I prefer to see things that I wouldn't pick at first but at least are consistent across projects.

packages/vite-plugin-cloudflare/src/runner/module-runner.ts
petebacondarwin: Leave it in for now. It could be some time before I get this nodejs stuff working properly... 😢

vicb: What would be super nice is to (create and?) link a bug for stuff like that

dario-piotrowicz: done 👍

packages/vite-plugin-cloudflare/src/workerd-custom-import.cts
dario-piotrowicz: I am quite sure it is a workerd issue

named exports imported from ESM can be found under `default`, but can also be found as named exports, that works in node.js but not in workerd that's the issue here

For example, the following code does work in node.js but would not work in workerd:
Screenshot 2024-10-09 at 14 35 51

dario-piotrowicz: oh wait... https://github.com/cloudflare/workerd/pull/2194 got merged... so I should double check, maybe this is actually no longer the case!

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: I've updated the code here 🙂 

I'll resolve this conversation if that's ok we can then follow up separately on formatting/linting 🙂

packages/vite-plugin-cloudflare/src/module-fallback.ts
dario-piotrowicz: done and done 😄👍

packages/vite-plugin-cloudflare/src/workerd-custom-import.cts
dario-piotrowicz: yeah It's no longer needed! 🚀 

Thanks for bringing this up, this allowed me to greatly simplify the code! 💪

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: it seems like in vite this whole builtin handling is done in a pretty hardcoded manner: 
https://github.com/vitejs/vite/blob/889bfc0ada6d6cd356bb7a92efdce96298f82fef/packages/vite/src/node/utils.ts#L95-L118

🤔

dario-piotrowicz: I am leaning into thinking that this needs to be hardcoded to a certain degree... maybe it is actually valid to put it as the last check in the method, after everything else has failed...

meaning that we resolve a `cloudflare:*` import as a builtin one only after everything else has failed, meaning that if the env consumer wants to polyfill the import or whatever they'd like we'd still enable that 🤔

jamesopstad: Good find. Would it make sense to get `cloudflare:` added to this list? It looks like the `isBuiltIn` function gets called by `fetchModule` - https://github.com/vitejs/vite/blob/889bfc0ada6d6cd356bb7a92efdce96298f82fef/packages/vite/src/node/ssr/fetchModule.ts#L32. I guess this still wouldn't work if `cloudflare:` is imported inside `node_modules`.

dario-piotrowicz: > Would it make sense to get `cloudflare:` added to this list?

I've asked this in the discord channel, I feel like that might not be the right approach, but it's worth double checking I guess 🙂

dario-piotrowicz: > I guess this still wouldn't work if `cloudflare:` is imported inside `node_modules`.

No this would not get triggered by code from `node_modules`, but it would still work totally fine since it runs outside of the vite module resolution

To be extra sure I've added tests for this: https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/15/commits/bddf610f7a509823dd134e988f0523ca7c3eee44

dario-piotrowicz: > I've asked this in the discord channel, I feel like that might not be the right approach, but it's worth double checking I guess 🙂

[They told me](https://discord.com/channels/804011606160703521/831456449632534538/1293625957994987551) that `cloudflare:` should not be included there, and that node, deno and bun are because apparently those are all platforms that can run a vite dev server 😮

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -110,37 +148,55 @@ export function cloudflare<
...
+									}
+									throw new Error(
+										`Unexpected Error, failed to get module: ${moduleId}`,
+									);
 								}
 							},
 						},
 					};
 				}),
+				unsafeUseModuleFallbackService: true,
```
petebacondarwin, line 198: Why is this twice, here and on line 140 above?

dario-piotrowicz: left over from refactoring, sorry 🙇

packages/vite-plugin-cloudflare/src/index.ts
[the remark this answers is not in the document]
dario-piotrowicz: sounds good 🙂👍 

if it's ok I'll do it as part of #18 so that I can also validate it as I add it 🙂

jamesopstad (approved):
This looks good to me. Once it's merged we can add the CI with the different operating systems and follow up on any issues that come up there.

packages/vite-plugin-cloudflare/src/module-fallback.ts
```diff
@@ -0,0 +1,384 @@
...
+
+async function collectModuleInfo(
+	moduleCode: string,
+	moduleFilePath: string,
+): Promise<CjsModuleInfo | EsmModuleInfo | JsonModuleInfo> {
+	if (moduleFilePath.endsWith('.json')) {
+		return { moduleType: 'json' };
+	}
+
+	if (!isCommonJS(moduleCode)) {
```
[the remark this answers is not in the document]
petebacondarwin: Sorry my logic was off. What I meant was:

```suggestion
	if (
		moduleFilePath.endsWith('.mjs') ||
		!(moduleFilePath.endsWith('.cjs') || isCommonJS(moduleCode))
	) {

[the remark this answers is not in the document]
dario-piotrowicz: ah ok 🙂👍

PR #18, opened 2024-10-08
add pull request checks github action and alongide related adjustments

dario-piotrowicz:


.github/workflows/pull-request-checks.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Check formatting
+        shell: bash
+        run: pnpm prettier:check
+
+      - name: Check types
+        shell: bash
+        run: pnpm types:check
+
+      - name: Run tests
+        if: matrix.os != 'windows-2022'
```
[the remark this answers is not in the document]
jamesopstad: Were you able to figure out what didn't work on Windows?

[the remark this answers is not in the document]
dario-piotrowicz: Not yet 😕 

I'm addressing the comments in my other PR and going to look more into this...

Although the windows error is pretty random/unclear to me: https://github.com/flarelabs-net/vite-plugin-cloudflare/actions/runs/11240997254/job/31251573045#step:8:23

(unsurprisingly is related to paths! 😓)

I am clueless when it comes to windows (and also don't have a windows machine to test things on) so if you have any insights/ideas I'm all ears!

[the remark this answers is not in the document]
dario-piotrowicz: PS: I am also very curious to merge my other PR and see if my new tests there (which worked in the old repo) do work here 🤔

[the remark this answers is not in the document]
jamesopstad: I wonder if we need to normalise the paths that are used for the workerd modules. In the `vitest-pool-workers` package they do this - https://github.com/cloudflare/workers-sdk/blob/4cd7ea9393fc9767024ad392b9c9bc7a42a9dd36/packages/vitest-pool-workers/src/pool/index.ts#L497.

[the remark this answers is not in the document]
dario-piotrowicz: ah no ok I also have to specify `contents`.... giving it another go 👍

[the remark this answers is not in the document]
dario-piotrowicz: should be all good now 👍

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -147,12 +157,16 @@ export function cloudflare<
 							{
 								type: 'ESModule',
 								path: runnerPath,
```
jamesopstad, line 159: Should all the paths be prefixed with `miniflareModulesRoot`?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -147,12 +157,16 @@ export function cloudflare<
...
+									fileURLToPath(new URL(runnerPath, import.meta.url)),
+									'utf8',
+								),
 							},
 							{
 								// we declare the workerd-custom-import as a CommonJS module, thanks to this
 								// require is made available in the module and we are able to handle cjs imports
 								type: 'CommonJS',
-								path: workerdCustomImportPath,
+								path: `${miniflareModulesRoot}${WORKERD_CUSTOM_IMPORT_PATH}`,
```
jamesopstad, line 169: Probably safer to use `path.join` here.

packages/vite-plugin-cloudflare/src/runner/module-runner.ts
```diff
@@ -18,10 +18,8 @@ export async function createModuleRunner(
 		throw new Error('Runner already initialized');
 	}
 
-	// we store the custom import file path in a variable to skip esbuild's import resolution
-	const workerdReqImport = '/__workerd-custom-import.cjs';
 	const { default: workerdCustomImport } = await (import(
-		workerdReqImport
+		`/${WORKERD_CUSTOM_IMPORT_PATH}`
```
jamesopstad, line 22: Is it OK to assume the `/` prefix here?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -147,12 +157,16 @@ export function cloudflare<
 							{
 								type: 'ESModule',
 								path: runnerPath,
+								contents: fs.readFileSync(
```
jamesopstad, line 160: It's kind of annoying that we have to read the module using `fs`. I saw this in the `vitest-pool-workers` package and was hoping it would work without it but it seems like it's the only way.

jamesopstad (commented):
Just a few comments/questions before we merge to check I understand this right.

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: this is a relative path, I think that only absolute ones need to have the `miniflareModulesRoot` prefix

packages/vite-plugin-cloudflare/src/runner/module-runner.ts
dario-piotrowicz: I think so, because we set it at the root in the miniflare constructor, and within miniflare we specify the path just with `/` (without the need to add the windows drive letter)

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: yeah I agree 😓

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: updated 🙂👍

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: so I might be wrong... let me double check...

packages/vite-plugin-cloudflare/src/runner/module-runner.ts
jamesopstad: Can you see if it works just using `WORKERD_CUSTOM_IMPORT_PATH` directly without adding `/` in front? That's what they do in vitest-pool-workers.

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: yeah I think it was probably wrong (but it didn't really matter as every user of the runner was in the same directory the runner is in) 👍

packages/vite-plugin-cloudflare/src/runner/module-runner.ts
dario-piotrowicz: not including the `/` does seem to break things:
Screenshot 2024-10-10 at 15 13 57

jamesopstad:
Do you think it's worth adding a `syncpack` check to ensure that dependency versions are consistent across the repo? They have a GitHub action - https://github.com/marketplace/actions/syncpack-synchronise-monorepo-dependency-versions. I think having it set to `version-mismatches: 'lint'` might be safer than automatic fixing.

jamesopstad:
@petebacondarwin Are you able to have a quick look at this PR? You probably have a better understanding of the logic around what does and doesn't work for path names.

dario-piotrowicz:
> Do you think it's worth adding a `syncpack` check to ensure that dependency versions are consistent across the repo? They have a GitHub action - https://github.com/marketplace/actions/syncpack-synchronise-monorepo-dependency-versions. I think having it set to `version-mismatches: 'lint'` might be safer than automatic fixing.

Now that you mention `syncpack`.... we're doing something differently in https://github.com/opennextjs/opennextjs-cloudflare and using pnpm catalogs instead: https://github.com/opennextjs/opennextjs-cloudflare/blob/450efc77395d46e05f0bf1aceaeb1c82625ff8ed/pnpm-workspace.yaml#L5

What if we removed `syncpack` and leaned on pnpm catalogs here as well instead? (btw that's also what `workers-sdk` is doing)

jamesopstad:
It would also be good to add the required pnpm version to the root `package.json` in that case.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -123,7 +138,7 @@ export function cloudflare<
 			const miniflare = new Miniflare({
 				workers: workers.map((workerOptions) => {
 					const wrappers = [
-						`import { createWorkerEntrypointWrapper } from '${runnerPath}';`,
+						`import { createWorkerEntrypointWrapper } from '${['.', ...rawRunnerPath].join('/')}';`,
```
petebacondarwin, line 141: Why the complexity around creating this runner path?

petebacondarwin (approved):

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: it's actually not!

it's a left over from when I was trying to get this to work, I've removed this unnecessary extra complexity

thanks for pointing it out 🙏

dario-piotrowicz: Sorry I spoke too soon... https://github.com/flarelabs-net/vite-plugin-cloudflare/actions/runs/11292722994/job/31409289435?pr=18#step:8:31

having the drive letter just seem to confuse workerd so this does need to be a standard path (which is built based on the fact that the wrapper module is also at the root)

jamesopstad: It would be really good to try and document the reasoning for what does and doesn't work with the path names. It all feels a bit like guesswork and I'm struggling to see the logic in it.

dario-piotrowicz: I've added a comment to it, please have a look 🙂

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -123,7 +141,10 @@ export function cloudflare<
 			const miniflare = new Miniflare({
 				workers: workers.map((workerOptions) => {
 					const wrappers = [
-						`import { createWorkerEntrypointWrapper } from '${runnerPath}';`,
+						// Note: this import relies on the `relativeRunnerPathSegments` array because using the full `runnerPath`
+						//       in windows would not work since `runnerPath` would start with `Z:` and workerd would
+						//       not know how to properly handle such path
+						`import { createWorkerEntrypointWrapper } from '${['.', ...relativeRunnerPathSegments].join('/')}';`,
```
jamesopstad, line 147: Sorry to be picky but can we put this path in a variable with the others. It doesn't feel right for the logic to live here.

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: Yeah by the way it is a bit messy, and I don't make things better by forgetting why I added something and just remembering when the windows check fails 😓 

Anyways, the whole issue here is that we're setting the extra windows prefix `Z:` and using potentially-windows paths (since we create them with `path.join`) and workerd does not know how to recognize such paths so in the code we pass to it we do need to use standard unix-like ones instead

hopefully that makes sense?

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Actually, don't worry about this now. It can be refactored later.

dario-piotrowicz: lol sorry I already did 😅

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: I think this is much clearer if we add the root at the point the module is provided to miniflare.
See https://github.com/cloudflare/workers-sdk/blob/1c9a9af8688eb0d8701184b5af847969fcf8024d/packages/vitest-pool-workers/src/pool/index.ts#L497-L509

That way it is easy to know that we haven't missed one.

jamesopstad (approved):

PR #23, opened 2024-10-09
Node.js compatibility (using prebundled unenv)

petebacondarwin:


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -5,16 +5,17 @@ import { unstable_getMiniflareWorkerOptions } from 'wrangler';
...
 import type { WorkerOptions } from 'miniflare';
 import type {
 	CloudflareEnvironmentOptions,
 	CloudflareDevEnvironment,
 } from './cloudflare-environment';
-import { getModuleFallbackHandler } from './module-fallback';
 import type { ResolveIdFunction } from './module-fallback';
 
-const wrapperPath = '__VITE_WRAPPER_PATH__';
+const wrapperPath = '__VITE_WORKER_ENTRY_POINT__';
```
petebacondarwin, line 18: I changed this simply because it makes more sense to the user when they see an error that contains this in the stack trace.

jamesopstad: Sure. One thing I've come to realise is that the word 'entrypoint' is reserved for individual exports (e.g. named entrypoints). I've started using the word 'entry' for the files instead so this could be `__VITE_WORKER_ENTRY__`.

package.json
```diff
@@ -13,5 +13,10 @@
 	"devDependencies": {
 		"prettier": "^3.3.3",
 		"syncpack": "^13.0.0"
+	},
+	"pnpm": {
+		"patchedDependencies": {
+			"unenv-nightly": "patches/unenv-nightly.patch"
```
petebacondarwin, line 19: I think I need to get a PR up for unenv - we can't live on this patch.

packages/vite-plugin-cloudflare/src/runner/module-runner.ts
```diff
@@ -5,11 +5,6 @@ import type { WrapperEnv } from './env';
 
 let moduleRunner: ModuleRunner;
 
-// TODO: node modules using process.env don't find `process` in the global scope for some reason
-//       for now we just create a `process` in the global scope but a proper solution needs to be
-//       implemented (see: https://github.com/flarelabs-net/vite-plugin-cloudflare/issues/22)
-(globalThis as Record<string, unknown>).process = { env: {} };
-
```
petebacondarwin, line 12: I moved this polyfill to the integration test that needs it instead of turning it on for everyone.

I tried turning on nodejs_compat for that `react` library but then it broke in other ways...

tests/integration/module-resolution/wrangler.toml
```diff
@@ -1 +1 @@
-compatibility_date = "2024-09-09"
+compatibility_date = "2024-09-01"
```
petebacondarwin, line 1: Pointless change, I will revert this.

packages/vite-plugin-cloudflare/src/module-fallback.ts
```diff
@@ -408,3 +423,22 @@ function isCommonJS(code: string): boolean {
...
+	// That function doesn't exist when run in `vm.runInThisContext()`
+	// i.e. inside a Vitest test.
+	const entryFile = fileURLToPath(
+		importMetaResolve(specifier, import.meta.url),
+	);
+	const transpiled = await build({
+		entryPoints: [entryFile],
+		bundle: true,
+		write: false,
+		format: 'cjs',
```
jamesopstad, line 438: Just for my understanding, why is it we bundle as cjs here?

tests/integration/module-resolution/index.ts
```diff
@@ -11,6 +11,10 @@ const allowedPaths = new Set([
 	'/third-party/slash-create',
 ]);
 
+// Needed by the react 3rd party library
+// Turning on nodejs_compat v2 causes the `slash-create` library to fail
```
jamesopstad, line 15: In theory, should code that runs successfully without `node_compat` also run with `node_compat` or are there scenarios where it's expected to break?

jamesopstad (approved):
Looks good to me. Just added a couple of questions for my own understanding. If possible, it would be good to add a couple of basic tests that verify things are broadly working as expected.

packages/vite-plugin-cloudflare/src/node-compat.ts
```diff
@@ -0,0 +1,92 @@
+import assert from 'node:assert';
```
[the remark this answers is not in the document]
petebacondarwin: Ah! I was wondering about that.

tests/integration/module-resolution/index.ts
petebacondarwin: I think the problem is to do with code "sniffing" what the platform is. If the library determines that it is running in Node.js (because of the reasonably high-fidelity polyfills provided by nodejs_compat) then it might run different code that doesn't actually work in workerd.

packages/vite-plugin-cloudflare/src/module-fallback.ts
petebacondarwin: In order for things like `require("node:buffer")` to work, I needed the module being uploaded to be of type `NodeJsCompatModule`, which required the code to be compiled to CommonJS.

petebacondarwin:
> If possible, it would be good to add a couple of basic tests that verify things are broadly working as expected.

Definitely! I should have left this in draft till that was added.
I was adding the tests when I got distracted by the logging and catalog bits.

tests/integration/module-resolution/index.ts
jamesopstad: I see now that `slash-create` is being imported from `slash-create/web`.

https://github.com/flarelabs-net/vite-plugin-cloudflare/blob/5893df31fedb6a8517cd02da6a714df37fb3be81/tests/integration/module-resolution/src/third-party/slash-create.ts#L1-L6

For Node, the standard import should be used so that explains the error.

packages/vite-plugin-cloudflare/src/node-compat.ts
```diff
@@ -0,0 +1,92 @@
...
+
+	// the mapping is a 2 item tuple, indicating a named export, made up of a module specifier and an export name.
+	const [moduleSpecifier, exportName] = globalInject;
+	return `import var_${globalName} from "${moduleSpecifier}";\nglobalThis.${globalName} = var_${globalName}.${exportName};`;
+}
+
+function deAliasing(alias: Record<string, string>) {
+	return (
+		[importSpecifier, modulePath]: [string, string],
+		index: number,
```
dario-piotrowicz, line 62:
```suggestion
```

it doesn't seem like `index` used?

packages/vite-plugin-cloudflare/src/node-compat.ts
```diff
@@ -0,0 +1,92 @@
...
+	): [string, string] => {
+		const breadcrumbs: (string | undefined)[] = [importSpecifier];
```
dario-piotrowicz, line 64: is this type correct/necessary?
```suggestion
		const breadcrumbs = [importSpecifier];
```

packages/vite-plugin-cloudflare/src/node-compat.ts
```diff
@@ -0,0 +1,92 @@
...
+		// the mapping is a simple string, indicating a default export, so the string is just the module specifier.
+		return `import var_${globalName} from "${moduleSpecifier}";globalThis.${globalName} = var_${globalName};`;
+	}
+
+	// the mapping is a 2 item tuple, indicating a named export, made up of a module specifier and an export name.
+	const [moduleSpecifier, exportName] = globalInject;
+	return `import var_${globalName} from "${moduleSpecifier}";\nglobalThis.${globalName} = var_${globalName}.${exportName};`;
+}
+
+function deAliasing(alias: Record<string, string>) {
```
dario-piotrowicz, line 59: __Very__ Nit: 
The function name is not imperative:
```suggestion
function deAlias(alias: Record<string, string>) {
```
or
```suggestion
function resolveAliasing(alias: Record<string, string>) {
```

packages/vite-plugin-cloudflare/src/module-fallback.ts
```diff
@@ -45,6 +48,18 @@ export function getModuleFallbackHandler(
 		const { resolveMethod, referrer, specifier, rawSpecifier } =
 			extractModuleFallbackValues(request);
 
+		if (rawSpecifier.startsWith('unenv/')) {
```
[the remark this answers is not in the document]
dario-piotrowicz: 🤔 I hope that we can get server-side pre-bundling working and actually avoid the need for this sort of thing across the board

Although unfortunately unenv and polyfills are a different story... might be worth exploring more 🤔

packages/vite-plugin-cloudflare/src/node-compat.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Get Miniflare modules and code snippets that are needed if running in Nodejs compat mode.
+ */
+export function getNodeCompatModules(
+	compatibilityDate: string | undefined,
+	compatibilityFlags: string[] = [],
+	modulesRoot: string,
+) {
+	if (getNodeCompat(compatibilityDate, compatibilityFlags).mode !== 'v2') {
```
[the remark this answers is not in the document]
petebacondarwin: Good shout. I definitely think we should avoid trying to support any other mode.

packages/vite-plugin-cloudflare/src/module-fallback.ts
```diff
@@ -408,3 +423,22 @@ function isCommonJS(code: string): boolean {
 
 	return false;
 }
+
+async function bundleDependency(specifier: string) {
+	// Can't use the standard `import.meta.resolve()` here.
+	// That function doesn't exist when run in `vm.runInThisContext()`
+	// i.e. inside a Vitest test.
```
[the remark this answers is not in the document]
petebacondarwin: In our own tests we run Vite inside a Vitest runner.
Such runners sometimes run in a sub-thread of Node.js and do not provide the `vm.runInThisContext()` function.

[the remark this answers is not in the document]
dario-piotrowicz: ah ok I see, thanks 🙂👍

PR #24, opened 2024-10-10
Add support for Durable Objects

### LINKED ISSUE #7: Support Durable Objects

jamesopstad:
resolves #7

This PR adds support for Durable Objects. They can be defined in the same worker or another worker (using `script_name`). RPC is also supported.


packages/vite-plugin-cloudflare/src/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+				const classNames = durableObjectClassNames[worker.name];
+				invariant(classNames, errorMessage(worker.name));
+
+				classNames.add(value);
+			} else if (typeof value === 'object') {
+				if (value.scriptName) {
+					const classNames = durableObjectClassNames[value.scriptName];
+					invariant(classNames, errorMessage(value.scriptName));
+
+					classNames.add(value.className);
```
dario-piotrowicz, line 62: here, are you taking the `scriptName` and using it as the `className`?

this doesn't seem correct to me... is it? 😮

jamesopstad: Here `scriptName` is the name of the worker where the DO is defined. So we get a reference to the `Set` of exports for that worker and then add the `className` to it.

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -13,6 +13,8 @@ interface WebSocketContainer {
 	webSocket?: WebSocket;
 }
 
+const webSocketError = 'The WebSocket is undefined';
```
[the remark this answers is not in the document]
jamesopstad: Fine with me.

packages/vite-plugin-cloudflare/src/utils.ts
jamesopstad: Maybe it's the naming that's confusing. I could call the variables `entryWorkerEntrypointNames` and `entryDurableObjectClassNames` to make it clearer that they are maps of entry files to exports.

dario-piotrowicz: yeah by debugging it I can see the logic there... but without context the variable names could be a bit confusing 🤔 

I personally would rename the variable and function and all that's related from a generic `DurableObjectClassNames` to something like `WorkersToDurableObjectClassNamesMap` or something like that, it is much more verbose but I personally feel like it makes things much much clearer here 🤔

jamesopstad:
Thanks Dario! I've made all the changes you suggested.

PR #26, opened 2024-10-14
Add state persistence

### LINKED ISSUE #21: Add options to persist state
Wrangler defaults to persisting state to `.wrangler/state/v3` in the config directory. This can be overridden with the `--persist-to` CLI flag. We have a few options:

1. Persist state to`.wrangler/state/v3` by default and add a `persistTo` option to the plugin config to override this.
2. Make state persistence opt-in (determined by whether the user provides the `persistTo` option in the plugin config).
3. Option 1 but also accept `false` as a value for `persistTo` in order to opt out.
4. Option 2 but also accept `true` as a value for `persistTo` in order to persist state to `.wrangler/state/v3`.

I'm most in favour of options 1 or 3 because they mirror the behaviour of `wrangler dev`. Users will still need to run Wrangler CLI commands when using the Vite plugin and the behaviour would be unexpected if they don't persist state to the same directory. It could be useful to opt out though (as in option 3), particularly for testing.

What do you think @IgorMinar @petebacondarwin @dario-piotrowicz?

jamesopstad:
resolves #21

This PR adds state persistence. By default, state is persisted to `.wrangler/state/v3` in the same directory as the Vite config. This matches the default behaviour of Wrangler. A new `persistTo` option has also been added to the plugin config. If a string is provided then this is used as the persistence directory. If `false` is provided then state is not persisted between runs of the dev server.

I have set `persistTo` to `false` in the integration tests.


.gitignore
```diff
@@ -4,3 +4,4 @@ dist
 .vscode
 .DS_Store
 **/*.tsbuildinfo
+.wrangler
```
dario-piotrowicz, line 7:
```suggestion
.wrangler

```

packages/vite-plugin-cloudflare/src/utils.ts
```diff
@@ -1,5 +1,27 @@
+import * as path from 'node:path';
 import { invariant } from './shared';
-import type { WorkerOptions } from 'miniflare';
+import type { SharedOptions, WorkerOptions } from 'miniflare';
+
+const DEFAULT_PERSIST_PATH = '.wrangler/state/v3';
```
dario-piotrowicz, line 5: Nit: I thought your preference here was to always rely on `path.join` instead of manually creating paths with `/`?

packages/vite-plugin-cloudflare/src/utils.ts
```diff
@@ -1,5 +1,27 @@
...
+
+export function getPersistence(
+	persistTo: string | false | undefined,
+	root: string,
+): Partial<SharedOptions> {
```
dario-piotrowicz, line 10: `Partial<SharedOptions>` seems a bit broad to me, what about narrowing it down?

```suggestion
type PersistOptions = Pick<
	SharedOptions,
	| 'cachePersist'
	| 'd1Persist'
	| 'durableObjectsPersist'
	| 'kvPersist'
	| 'r2Persist'
>;

export function getPersistence(
	persistTo: string | false | undefined,
	root: string,
): Partial<PersistOptions> {
```

packages/vite-plugin-cloudflare/src/utils.ts
```diff
@@ -1,5 +1,27 @@
...
+	if (persistTo === false) {
+		return {};
```
dario-piotrowicz, line 12: sorry I don't fully remember Miniflare's default behavior, are you sure this does what we expect to? or should we clearly set the various persistance fields to `false`?

packages/vite-plugin-cloudflare/src/utils.ts
jamesopstad: I think it's fine if the paths are static. I just think if you're joining paths then you might as well use `path.join` rather than template strings. I can use `path.join` if you prefer though.

packages/vite-plugin-cloudflare/src/utils.ts
jamesopstad: This does do what we expect because `undefined` is equivalent to `false` here.

https://github.com/cloudflare/workers-sdk/tree/main/packages/miniflare#type-persistence

I guess you could argue that setting each value to `false` explicitly is more resilient if this behaviour ever changed. What do you think?

.gitignore
dario-piotrowicz: adding an extra newline 😅 

Screenshot 2024-10-14 at 14 29 28

As I believe that having that is a best practice?
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

packages/vite-plugin-cloudflare/src/utils.ts
dario-piotrowicz: I see... I don't really have a strong preference, I just tend to prefer consistency across a codebase (even if it can make some lines a bit more verbose than necessary? 🤔)

packages/vite-plugin-cloudflare/src/utils.ts
dario-piotrowicz: Ah ok, thanks for checking that! all good then 😄 

> I guess you could argue that setting each value to false explicitly is more resilient if this behaviour ever changed. What do you think?

mh.... yeah I do think that that would be more resilient, but it would also be more verbose since you'd have to add all the various fields here 😕 (without an immediate benefit to back this up)

So if you do want to me more explicit that's fine by me, but I'm also ok if we want to have more concise code, I'm happy with whatever you prefer 🙂 (PS: even if miniflare's default behavior were to change and this would break it would be unlikely to create much damage and the fix would be pretty trivial)

jamesopstad: OK, sure. I've made the changes. As I've used the `PersistOptions` type without `Partial` we'd get a type error if Miniflare stopped accepting `undefined` for persist values. I think that's fine without the extra explicitness.

dario-piotrowicz (approved):
Looks great to me 😄

PR #32, opened 2024-10-15
[DO NOT MERGE] pre-bundling example (POC)

dario-piotrowicz:
> [!WARNING]
> This PR is not really for merging but more just to share the state of things with the prebundling investigation and gather feeback/start a conversation

> [!Note]
> When/if https://github.com/vitejs/vite/pull/18358 gets merged and released I'll probably extract what I did here in our tests and do try to remove as much module fallback logic as we can (Note: tests on users trying to opt-out of pre-bundling should be added as well)

### DIFF
--- a/examples/worker/src/index.ts
+++ b/examples/worker/src/index.ts
@@ -1,7 +1,12 @@
 import { OK } from 'zod';
+import React, { version as ReactVersion } from 'react';
 
 export default {
 	async fetch() {
-		return new Response('Hello World! ===> ' + OK.name);
+		return Response.json({
+			'zod.OK.name': OK.name,
+			'typeof React': typeof React,
+			'react version': ReactVersion,
+		});
 	},
 };
--- a/examples/worker/src/index.ts
+++ b/examples/worker/src/index.ts
@@ -1,5 +1,7 @@
+import { OK } from 'zod';
+
 export default {
 	async fetch() {
-		return new Response('Hello World!');
+		return new Response('Hello World! ===> ' + OK.name);
 	},
 };
--- a/examples/worker/src/index.ts
+++ b/examples/worker/src/index.ts
@@ -1,17 +1,18 @@
-import { OK } from 'zod';
-import React, { version as ReactVersion } from 'react';
+import { maxLength } from '@cloudflare-dev-module-resolution/builtins/cjs';
 import { RPCErrorCodes, Utils } from 'discord-api-types/v10';
+import React, { version as ReactVersion } from 'react';
 import { Collection, SlashCreator, VERSION } from 'slash-create/web';
+import { OK } from 'zod';
 
 const slashCreatorInstance = new SlashCreator({
 	applicationID: 'xxx',
 });
 
 const myCollection = new Collection([['a number', 54321]]);
-
 export default {
 	async fetch() {
 		return Response.json({
+			userAgent: navigator.userAgent,
 			'zod.OK.name': OK.name,
 			'typeof React': typeof React,
 			'react version': ReactVersion,
@@ -24,6 +25,7 @@ export default {
 			),
 			'(discord-api-types/v10) RPCErrorCodes.InvalidUser':
 				RPCErrorCodes.InvalidUser,
+			'node:buffer MAX_LENGTH': maxLength,
 		});
 	},
 };
--- a/examples/worker/src/index.ts
+++ b/examples/worker/src/index.ts
@@ -1,5 +1,13 @@
 import { OK } from 'zod';
 import React, { version as ReactVersion } from 'react';
+import { RPCErrorCodes, Utils } from 'discord-api-types/v10';
+import { Collection, SlashCreator, VERSION } from 'slash-create/web';
+
+const slashCreatorInstance = new SlashCreator({
+	applicationID: 'xxx',
+});
+
+const myCollection = new Collection([['a number', 54321]]);
 
 export default {
 	async fetch() {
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -13,5 +13,10 @@
 	"devDependencies": {
 		"prettier": "^3.3.3",
 		"syncpack": "^13.0.0"
+	},
+	"pnpm": {
+		"overrides": {
+			"vite": "/Users/dario/Repos/my-repos/vite/packages/vite"
```
dario-piotrowicz, line 19: Naturally this only works locally, I will push my stuff on my vite fork in a bit 🙂

packages/vite-plugin-cloudflare/src/module-fallback.ts
```diff
@@ -28,11 +28,7 @@ export function getResolveId(
 	});
 
 	return function resolveId(id, importer, options) {
-		const resolveMethod = options?.resolveMethod ?? 'import';
-		const resolveIdFn =
-			resolveMethod === 'import' ? esmResolveId : cjsResolveId;
-
-		return resolveIdFn(devEnvironment, id, importer);
+		throw new Error('Not implemented');
```
dario-piotrowicz, line 31: The idea is that this function wouldn't actually be needed anymore

Without my patch this change would result in:
Screenshot 2024-10-15 at 15 42 18

But with my patch `zod` actually gets correctly resolved:
Screenshot 2024-10-15 at 15 42 52

dario-piotrowicz: Note: I am throwing inside `resolveId` just as an example, but this basically indicates that most if not all of the module fallback service would not be needed anymore I think

dario-piotrowicz:
#### Question 
developers/plugins can always opt out of pre-bundling via `optimizeDeps.exclude` ([docs](https://vite.dev/config/dep-optimization-options.html#optimizedeps-exclude)) or `optimizeDeps.noDiscovery: true` ([docs](https://vite.dev/config/dep-optimization-options.html#optimizedeps-disabled))... if we actually went with prebundling, where does that leave us?

do we simply not support such options?

or do we support them meaning that we would have both pre-bundling and a potential fallback service fallback?
(basically recommending the more stable/robust pre-bundling but not forcing it)
(this would non-trivially increase maintenance costs...)

package.json
jamesopstad: It's great this requires so few changes to Vite! Is the idea that this is something you could opt in to in the environment config somehow (and we could set this option for the environments created by our plugin)?

jamesopstad:
> #### Question
> developers/plugins can always opt out of pre-bundling via `optimizeDeps.exclude` ([docs](https://vite.dev/config/dep-optimization-options.html#optimizedeps-exclude)) or `optimizeDeps.noDiscovery: true` ([docs](https://vite.dev/config/dep-optimization-options.html#optimizedeps-disabled))... if we actually went with prebundling, where does that leave us?
> 
> do we simply not support such options?
> 
> or do we support them meaning that we would have both pre-bundling and a potential fallback service fallback? (basically recommending the more stable/robust pre-bundling but not forcing it) (this would non-trivially increase maintenance costs...)

I think it would be best to just not support these options if we go this route. In my opinion the maintenance burden of keeping the unbundled behaviour alongside this would be too high.

package.json
dario-piotrowicz: > Is the idea that this is something you could opt in to in the environment config somehow (and we could set this option for the environments created by our plugin)?

Yes and yes, it's practically opt-in and we could set this in our plugin (this could still be overridden by the plugin's consumer)

To opt-in you simply need to let vite know where it should start its pre-bundling discovering crawling, and that's actually what I added here: https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/32/files#diff-ccdf400ac3f84526b65f252b32659a7a52e1bfbc70d008e1baf4f9a98c87b9f3R129-R134

jamesopstad: Ah, OK. I wasn't sure if interpreting the existing `optimizeDeps.entries` option in this way would be a breaking change. What happens in current Vite if you provide this option for a custom environment?

examples/worker/src/index.ts
```diff
@@ -1,5 +1,31 @@
...
+			'(slash-create/web) slashCreatorInstance is instance of SlashCreator':
+				slashCreatorInstance instanceof SlashCreator,
+			'(slash-create/web) myCollection.random()': myCollection.random(),
+			'(discord-api-types/v10) Utils.isLinkButton({})': Utils.isLinkButton(
+				{} as any,
+			),
+			'(discord-api-types/v10) RPCErrorCodes.InvalidUser':
+				RPCErrorCodes.InvalidUser,
+			'node:buffer MAX_LENGTH': maxLength,
+		});
```
dario-piotrowicz, line 29: This is the result:
Screenshot 2024-10-22 at 11 40 30

package.json
dario-piotrowicz: The above is no longer applicable since the way the ssr deps crawling is triggered has changed, now to enable it you need to set `noDiscovery: false` alongside the `optimizeDeps.entries` or `rollupOptions.input` values

(also `resolve.noExternal: true` to actually make sure that the dependencies do get pre-bundled)

So there is no breaking change concern 🙂

examples/worker/vite.config.ts
```diff
@@ -2,12 +2,23 @@ import { cloudflare } from '@flarelabs-net/vite-plugin-cloudflare';
...
 		cloudflare({
 			workers: {
 				worker: {
 					main: './src/index.ts',
 					wranglerConfig: './src/wrangler.toml',
+					overrides: {
+						// Note: this sets esbuild's platform to node, allowing it to accept
+						//       imports such as module, node:module, buffer, node:buffer, etc...
+						webCompatible: false,
+					},
```
dario-piotrowicz, line 21: TODO: if a worker's config includes `nodejs_compat` then we want to set `webCompatible` to false in our plugin... I think?

dario-piotrowicz: Done in: https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/32/commits/1d2ea14f10265041c324b93aa071979cb07c139c

dario-piotrowicz:
Closing in favour of #37

PR #33, opened 2024-10-15
fix: do not show unnecessary logs and make sure they are strings not …

petebacondarwin:
…buffers


jamesopstad (approved):

examples/worker/src/index.ts
```diff
@@ -1,5 +1,6 @@
 export default {
 	async fetch() {
+		throw new Error('xxx');
```
dario-piotrowicz, line 3: is this intentional? 😅

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -159,8 +159,11 @@ export function getMiniflareOptions(
 	return {
 		log: logger,
 		handleRuntimeStdio(stdout, stderr) {
-			stdout.forEach((data) => logger.info(data));
-			stderr.forEach((error) => logger.error(error));
+			const decoder = new TextDecoder();
+			stdout.forEach((data) => logger.info(decoder.decode(data)));
+			stderr.forEach((error) =>
+				logger.logWithLevel(LogLevel.ERROR, decoder.decode(error)),
+			);
```
dario-piotrowicz, line 166: just out curiosity, is the `TextDecoder` necessary? yesterday I tried with `data.toString()` and `error.toString()` and that seemed to do the job 🤔

dario-piotrowicz (approved):
Looks good to me 🙂 

Awesome to also see testing for this 🤩

examples/worker/src/index.ts
petebacondarwin: Oops!

packages/vite-plugin-cloudflare/src/miniflare-options.ts
petebacondarwin: This is just how I have always converted bytes to strings.

PR #35, opened 2024-10-16
Improve error handling for Miniflare errors

jamesopstad:
This improves the error handling on wrangler config changes so that the Vite dev server is not killed when Miniflare encounters an error. Instead, the error is captured and displayed in the browser overlay on the next request. When the error is resolved, the browser is programatically refreshed.

I tried to implement the same behaviour for initial startup but this won't work as the environments can't be created without Miniflare having successfully initialised.


petebacondarwin (approved):
It would be great to find a way to test this sort of thing....

jamesopstad:
> It would be great to find a way to test this sort of thing....

Yes, any ideas how we can write tests that edit configuration files?

I'll merge for now but will have a think about this.

PR #37, opened 2024-10-23
completely remove module-fallback logic (and rely on pre-bundling)

dario-piotrowicz:


package.json
```diff
@@ -14,5 +14,10 @@
 		"@ianvs/prettier-plugin-sort-imports": "^4.3.1",
 		"prettier": "^3.3.3",
 		"prettier-plugin-packagejson": "^2.5.3"
+	},
+	"pnpm": {
+		"overrides": {
+			"vite": "https://pkg.pr.new/vite@main"
```
dario-piotrowicz, line 20: Hopefully this is fine, if not we can wait to merge this until the next vite 6 beta release

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -1,4 +1,5 @@
 import * as vite from 'vite';
+import { nodeBuiltIns } from './node-builtin-modules';
```
petebacondarwin, line 2: Why not use:

```
import { builtinModules } = "node:modules";

const nodeBuiltIns = builtinModules.concat(builtinModules.map((m) => `node:${m}`);
```

dario-piotrowicz: What do you mean with `import { builtinModules } = "node:modules";`? 😮 

if you mean just exporting the non-prefixed modules in `node-builtin-modules` and do the prefixing using `concat` that sounds find by me 🙂

dario-piotrowicz: @petebacondarwin is this what you meant? 🙂  https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/37/commits/7aefe02ec78409c19893a2a52b03763c40064ea7

petebacondarwin: Sorry there was a typo in my suggestion. This is my suggestion:

```suggestion
import { builtinModules } from "node:modules";

const nodeBuiltIns = builtinModules.concat(builtinModules.map((m) => `node:${m}`);
```

dario-piotrowicz: Ah ok awesome! I wasn't aware of this API, I've applied the change, thanks! 😄

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -145,16 +144,6 @@ export function getMiniflareOptions(
...
-	// TODO: we only have a single module resolution strategy shared across all workers
-	//       (generated using the first worker's dev environment)
-	//       we should investigate and ideally have potential different resolutions per worker
-	//       see: https://github.com/flarelabs-net/vite-plugin-cloudflare/issues/19
-	const firstWorkerName = workers[0]?.name;
-	invariant(firstWorkerName, 'First worker name not found');
-	const devEnvironment = viteDevServer.environments[firstWorkerName];
-	invariant(devEnvironment, 'First worker dev environment not found');
-	const resolveId = getResolveId(viteConfig, devEnvironment);
-
```
dario-piotrowicz, line 157: This issue went away by itself 🎉 

To tweak the module resolution of a specific worker all you need to do is apply appropriate `overrides` on that worker 🙂

tests/integration/module-resolution/module-resolution.test.ts
```diff
@@ -89,6 +69,22 @@ describe('module resolution', async () => {
 		});
 	});
 
+	test('node built-ins (both from userland and external dependencies)', async () => {
```
petebacondarwin, line 72: So really node built-ins should fail unless we have nodejs_compat turned on.
Not sure, therefore, how I feel about this test...

petebacondarwin: Oh but I see you did turn it on in the wrangler.toml...
Which means that the non-polyfilled modules are available.

petebacondarwin (approved):

PR #39, opened 2024-10-23
pre-bundling follow ups

petebacondarwin:
Now that we are prebundling, and can turn on npdejs_compat, some of the hacks from before are no longer needed.

- remove the fake process global
- remove the workerd custom import module
- remove the unsafeUseModuleFallbackService property
- consolidate environment options creation
- move vite override to the workspace catalog

### DIFF
--- a/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
+++ b/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
@@ -1,4 +1,5 @@
 import { builtinModules } from 'node:module';
+import * as path from 'node:path';
 import * as vite from 'vite';
 import { INIT_PATH, invariant, UNKNOWN_HOST } from './shared';
 import type { NormalizedPluginConfig, WorkerOptions } from './plugin-config';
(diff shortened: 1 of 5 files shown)


pnpm-workspace.yaml
```diff
@@ -6,6 +6,6 @@ packages:
 catalog:
   '@cloudflare/workers-types': ^4.20240919.0
   'typescript': ^5.6.2
-  'vite': '6.0.0-beta.2'
+  'vite': "https://pkg.pr.new/vite@main"
```
dario-piotrowicz, line 9: Instead of using the main prerelease we can now use the latest 6 beta release: [`6.0.0-beta.4`](https://github.com/vitejs/vite/blob/v6.0.0-beta.4/packages/vite/CHANGELOG.md#600-beta4-2024-10-23)

or we can keep using the main prerelease to try to closely keep up to date with vite, whatever you guys prefer is fine by me

petebacondarwin: Cool! Yes, let's go with beta.4

petebacondarwin:
```suggestion
  'vite': '6.0.0-beta.4'
```

dario-piotrowicz (approved):
Looks good to me 😄 

Thanks for addressing the spots I missed with my prebundling PR 🙏

jamesopstad:
There was a regression in this PR that means you now can't override the output directory for a Worker in the Vite config (this was why I used `configEnvironment` before, which has been removed). I've added a new issue - https://github.com/flarelabs-net/vite-plugin-cloudflare/issues/45.

PR #43, opened 2024-10-28
[WIP - DO NOT MERGE] add validation for pre-bundling opt-out

dario-piotrowicz:


packages/vite-plugin-cloudflare/src/runner/module-runner.ts
```diff
@@ -69,7 +77,10 @@ export async function createModuleRunner(
 				Object.freeze(context.__vite_ssr_exports__);
 			},
 			async runExternalModule(filepath) {
-				filepath = filepath.replace(/^file:\/\//, '');
+				return new Error(
+					`[Error] Trying to import external module (only internal modules are allowed): ${module.id}`,
+				);
```
dario-piotrowicz, line 82: This does not work 😕 

I think the Vite module runner is swallowing the error at some point before it's returned to the user

I'm thinking of parking this effort for now and come back when we do have inspectors working, as that would really really simplify the investigation needed here

playground/module-resolution/__tests__/module-resolution.spec.ts
```diff
@@ -3,11 +3,17 @@ import {
...
 	serverLogs,
+	viteTestUrl,
 } from '../../__test-utils__';
 
 describe.runIf(!isBuild)('module resolution', async () => {
-	afterAll(() => expect(serverLogs.errors).toEqual([]));
+	afterAll(() => {
+		const isNotUnconfiguredAliasError = (error: string) =>
+			!error.includes('@alias/unconfigured');
+		expect(serverLogs.errors.filter(isNotUnconfiguredAliasError)).toEqual([]);
```
petebacondarwin, line 15: Is doing expectations in `afterAll()` an anti-pattern? How easy will it be to see which test caused the problems if the expectation fails?

playground/module-resolution/__tests__/module-resolution.spec.ts
```diff
@@ -106,5 +112,13 @@ describe.runIf(!isBuild)('module resolution', async () => {
 			const result = await getTextResponse('/@alias/test');
 			expect(result).toBe('OK!');
 		});
+
+		test('imports from a non configured aliased package', async () => {
```
petebacondarwin, line 116: What does "non configured" actually mean?

playground/module-resolution/vite.config-no-prebundling.ts
```diff
@@ -0,0 +1,42 @@
...
+									'react',
+								],
+							},
+						},
+						resolve: {
+							external: ['@cloudflare-dev-module-resolution/requires/ext'],
+						},
+					},
+				},
+			},
```
petebacondarwin, line 37: Can you put a comment explaining what is going on here?

playground/module-resolution/__tests__/module-resolution.spec.ts
petebacondarwin: I think you mean non-aliased, right?

petebacondarwin (approved):

playground/module-resolution/vite.config-no-prebundling.ts
dario-piotrowicz: sure 🙂👍

dario-piotrowicz: I've added some comments, please have a look 🙏

jamesopstad:
No need to wait on me for this one if you want to merge before I get to it.

petebacondarwin (approved):
Love the comments! Thanks

playground/module-resolution/__tests__/module-resolution.spec.ts
dario-piotrowicz: I've updated the code and test to clarify this 🙂 

Having it as an alias, which just so happen not to be correctly configured is probably the wrong mindset, I've updated it to be a non-existing package instead, so that the testing is regarding a user error in which a non existing package gets imported

Please have another look 🙂

PR #44, opened 2024-10-30
make sure that the root `check:types` script includes the playground

dario-piotrowicz:
Currently if you run `check:types` at the root of the repo the playground and its apps aren't actually getting checked, I am addressing that here

Sorry I missed this in my refactoring PR 😓

(I've been just manually checking the types via Vscode)

### DIFF
--- a/playground/vitest-setup.ts
+++ b/playground/vitest-setup.ts
@@ -99,6 +99,7 @@ beforeAll(async (s) => {
 	browser = await chromium.connect(wsEndpoint);
 	page = await browser.newPage();
 
+	// @ts-ignore
 	const globalConsole = global.console;
 	const warn = globalConsole.warn;
 	globalConsole.warn = (msg: string, ...args: unknown[]) => {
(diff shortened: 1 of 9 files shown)


playground/vitest-setup.ts
```diff
@@ -99,6 +99,7 @@ beforeAll(async (s) => {
 	browser = await chromium.connect(wsEndpoint);
 	page = await browser.newPage();
 
+	// @ts-ignore
```
dario-piotrowicz, line 102: This is needed because unfortunately this file is getting pulled in from the various playground apps (since they import from `../../__test-utils__` which imports from this file) and they error saying:
```
error TS2339: Property 'console' does not exist on type 'typeof globalThis'.
```

This seems to be caused by the use of the workers types... I am not really 100% sure what the problem is, I'm trying to figure that out 😕

dario-piotrowicz: This seemed to do the trick: https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/44/commits/8013a3c8b5d4c5eec56dea85fcf02c87822aebe0 🤷

dario-piotrowicz: I think we're missing a definition for `globalThis` in the workers types? 😕

jamesopstad (approved):
Thanks. Sorry I missed that too.

PR #50, opened 2024-11-05
bump vite to `6.0.0-beta.8`

dario-piotrowicz:
> [!Warning]
> This PR is blocked by this issue: https://github.com/vitejs/vite/pull/18395/files#r1832458011
> Fixed by: https://github.com/vitejs/vite/pull/18611 (this will apply only in the next Vite beta 10 release though)

### DIFF
--- a/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
+++ b/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
@@ -166,6 +166,7 @@ export function createCloudflareEnvironmentOptions(
 					...builtinModules.concat(builtinModules.map((m) => `node:${m}`)),
 				],
 				esbuildOptions: {
+					platform: 'neutral',
 					resolveExtensions: [
 						'.mjs',
 						'.js',
--- a/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
+++ b/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
@@ -135,6 +135,8 @@ export function createCloudflareEnvironmentOptions(
 				// Note: in order for ssr pre-bundling to take effect we need to ask vite to treat all
 				//       dependencies as not external
 				noExternal: true,
+				// We want to use `workerd` package exports if available (e.g. for postgres).
+				conditions: ['workerd', 'module', 'browser', 'development|production'],
 			},
 			dev: {
 				createEnvironment(name, config) {
--- a/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
+++ b/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
@@ -140,28 +140,6 @@ export function createCloudflareEnvironmentOptions(
 				createEnvironment(name, config) {
 					return new CloudflareDevEnvironment(name, config);
 				},
-				optimizeDeps: {
-					// Note: ssr pre-bundling is opt-in, and we need to enabled it by setting noDiscovery to false
-					noDiscovery: false,
-					exclude: [
-						...cloudflareBuiltInModules,
-						...builtinModules.concat(builtinModules.map((m) => `node:${m}`)),
-					],
-					esbuildOptions: {
-						resolveExtensions: [
-							'.mjs',
-							'.js',
-							'.mts',
-							'.ts',
-							'.jsx',
-							'.tsx',
-							'.json',
-							'.cjs',
-							'.cts',
-							'.ctx',
-						],
-					},
-				},
 			},
 			build: {
 				createEnvironment(name, config) {
@@ -178,7 +156,29 @@ export function createCloudflareEnvironmentOptions(
 					external: [...cloudflareBuiltInModules],
 				},
 			},
-			webCompatible: true,
+			optimizeDeps: {
+				// Note: ssr pre-bundling is opt-in, and we need to enabled it by setting noDiscovery to false
+				noDiscovery: false,
+				exclude: [
+					...cloudflareBuiltInModules,
+					...builtinModules.concat(builtinModules.map((m) => `node:${m}`)),
+				],
+				esbuildOptions: {
+					resolveExtensions: [
+						'.mjs',
+						'.js',
+						'.mts',
+						'.ts',
+						'.jsx',
+						'.tsx',
+						'.json',
+						'.cjs',
+						'.cts',
+						'.ctx',
+					],
+				},
+			},
+			keepProcessEnv: true,
 		} satisfies vite.EnvironmentOptions,
 		options.overrides ?? {},
 	);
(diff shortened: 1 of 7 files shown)


packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -140,28 +140,6 @@ export function createCloudflareEnvironmentOptions(
 				createEnvironment(name, config) {
 					return new CloudflareDevEnvironment(name, config);
 				},
-				optimizeDeps: {
```
dario-piotrowicz, line 143: Vite's moved this option to the top level: https://github.com/vitejs/vite/pull/18465

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -178,7 +156,29 @@ export function createCloudflareEnvironmentOptions(
 					external: [...cloudflareBuiltInModules],
 				},
 			},
-			webCompatible: true,
```
dario-piotrowicz, line 181: They've also removed `webCompatible` and kind of replaced it with `keepProcessEnv` (https://github.com/vitejs/vite/pull/18395#discussion_r1820503185)

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -178,7 +156,29 @@ export function createCloudflareEnvironmentOptions(
...
+						'.jsx',
+						'.tsx',
+						'.json',
+						'.cjs',
+						'.cts',
+						'.ctx',
+					],
+				},
+			},
+			keepProcessEnv: true,
```
dario-piotrowicz, line 181: as far as I understand `keepProcessEnv: true` basically asks Vite not to mess with `process.env` which I think is something that we want ([source](https://github.com/vitejs/vite/blob/ce0eec91c9b5a91c52bf5b9bae6173d35f6e1649/packages/vite/src/node/config.ts#L246-L250))

jamesopstad:
I think it's best to hold off upgrading until #49 is merged.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -33,8 +33,6 @@ export function cloudflare<T extends Record<string, WorkerOptions>>(
 			return {
 				resolve: {
 					alias: getNodeCompatAliases(),
-					// We want to use `workerd` package exports if available (e.g. for postgres).
-					conditions: ['workerd'],
```
dario-piotrowicz, line 37: @petebacondarwin I needed to move this to the environment config since `conditions` is no longer inherited by environments ([source change](https://github.com/vitejs/vite/commit/d002e7d05a0f23110f9185b39222819bcdfffc16#diff-11e17761d4ecfee8f8fde15c6d79b7bc0260176396a30dfd8e6f6bbaf5af4745R682)) I hope this is fine, was there a particular reason why you added the `conditions` here?

petebacondarwin: This was the place I found that worked. But if it works in the environment config that is also cool. 
Does that cover builds too?

petebacondarwin:
Actually the CI jobs are not happy at all...

dario-piotrowicz:
> Actually the CI jobs are not happy at all...

yes, things only pass with my local version of vite (unless I do apply a pnpm patch)

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: > Does that cover builds too?

It is a top-level configuration so I would definitely say yes 🤔

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -178,7 +158,30 @@ export function createCloudflareEnvironmentOptions(
...
-			webCompatible: true,
+			optimizeDeps: {
+				// Note: ssr pre-bundling is opt-in, and we need to enabled it by setting noDiscovery to false
+				noDiscovery: false,
+				exclude: [
+					...cloudflareBuiltInModules,
+					...builtinModules.concat(builtinModules.map((m) => `node:${m}`)),
+				],
+				esbuildOptions: {
+					platform: 'neutral',
```
dario-piotrowicz, line 169: @petebacondarwin @jamesopstad I put `neutral` as that seemed the preference in chat? do you agree? or do you think that `browser` would be better?

jamesopstad: I think we should go with `neutral` as it seems more appropriate and then we can always change to `browser` later if needed.

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
jamesopstad: Interesting. Perhaps using `false` so that it replaces it with an empty object would be good when `nodejs_compat` is not enabled?

jamesopstad (approved):
Looks good to me!

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
dario-piotrowicz: sounds good to me 🙂

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
dario-piotrowicz: I don't have a super strong opinion on this one

I think I am a bit more inclined to say that `keepProcessEnv: true` is always the best in our case so that we completely leave the handling/initialization of `process.env` up to workerd

but again I am fine with either 🙂  (it should also be simple to revert this back later, although having to check if nodejs_compat is set is sound to force us to move this code around unfortunately I think (as I did here: https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/52/files#diff-0217c39a7be6fcdb410ac8dfea2ae2998b882a1aaba3f2e83ff2831a0beee7bfR71-R73))

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
petebacondarwin: Also neutral adds in the fewest extra behaviours, which we can (mostly) add in manually if needed. The other modes require us to remove behaviours, which is generally more tricky.

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
petebacondarwin: I agree that keeping it set to true for now seems the best approach.

petebacondarwin (approved):

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
dario-piotrowicz: ah ok, let's go with that for now then (and we can always revisit later on if needed) 🙂

PR #52, opened 2024-11-05
use the new `resolve.builtins` vite config to specify the cloudflare and node builtins

### LINKED ISSUE #46: Investigate the `fetchModule`'s `cloudflare:` catch
In fetchModule, if there is an error we have this catch logic:
https://github.com/flarelabs-net/vite-plugin-cloudflare/blob/7adf439fbd5e187c7419fe5686ed653191363d87/packages/vite-plugin-cloudflare/src/miniflare-options.ts#L230-L237

that externalizes `cloudflare:*` modules

Is this really correct?

Shouldn't `cloudflare:*` modules be externalized already?
https://github.com/flarelabs-net/vite-plugin-cloudflare/blob/7adf439fbd5e187c7419fe5686ed653191363d87/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts#L171-L175
(this probably only applies to build 🤔)

I feel like we might be missing some config here, and we probably should not get `cloudflare:*` in the fetchModule 🤔 
(or if this is ok it should be clear as to why we do need this ad-hoc module handling)

dario-piotrowicz:
resolves #46

paired with this PR I created in the Vite repo: https://github.com/vitejs/vite/pull/18584

> [!Note]
> The changes in https://github.com/vitejs/vite/pull/18584 are most likely going to make it into vite around Jannuary ([discord message](https://discord.com/channels/804011606160703521/831456449632534538/1316002878761996288)) so this PR will be blocked until then

### DIFF
--- a/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
+++ b/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
@@ -1,7 +1,5 @@
 import assert from 'node:assert';
-import { builtinModules } from 'node:module';
 import * as vite from 'vite';
-import { getNodeCompatExternals } from './node-js-compat';
 import { INIT_PATH, UNKNOWN_HOST } from './shared';
 import { toMiniflareRequest } from './utils';
 import type { ResolvedPluginConfig, WorkerConfig } from './plugin-config';
@@ -153,7 +151,6 @@ export function createCloudflareEnvironmentOptions(
 				//       dev pre-bundling crawling (were we not to set this input field we'd have to appropriately set
 				//       optimizeDeps.entries in the dev config)
 				input: workerConfig.main,
-				external: [...getNodeCompatExternals()],
 			},
 		},
 		optimizeDeps: {
--- a/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
+++ b/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
@@ -133,6 +133,7 @@ export function createCloudflareEnvironmentOptions(
 			noExternal: true,
 			// We want to use `workerd` package exports if available (e.g. for postgres).
 			conditions: ['workerd', 'module', 'browser', 'development|production'],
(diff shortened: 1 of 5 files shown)


pnpm-workspace.yaml
```diff
@@ -6,7 +6,7 @@ packages:
 catalog:
   '@cloudflare/workers-types': ^4.20241112.0
   'typescript': ^5.6.2
-  'vite': '6.0.0-beta.10'
+  'vite': '/Users/dario/Repos/my-repos/vite/packages/vite'
```
dario-piotrowicz, line 9: Clearly this PR won't be ready to merge until the version of vite with the `resolve.builtins` config is (hopefully) released

dario-piotrowicz, line 9: Clearly this PR won't be ready to merge until the version of vite with the resolve.builtins config is (hopefully) released

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -155,17 +155,12 @@ export function createCloudflareEnvironmentOptions(
 					input: userConfig.root
 						? path.resolve(userConfig.root, options.main)
 						: path.resolve(options.main),
-					external: [...cloudflareBuiltInModules, ...getNodeCompatExternals()],
+					external: [...getNodeCompatExternals()],
```
dario-piotrowicz, line 158: @petebacondarwin we set these node externals unconditionally even if the worker is not under nodejs_compat, is that ok?

Or should we do it conditionally? (similarly to the [`addNodeBuiltinsIfNeeded` function](https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/52/files#diff-0217c39a7be6fcdb410ac8dfea2ae2998b882a1aaba3f2e83ff2831a0beee7bfR222) that I'm adding here)

petebacondarwin: We can do it conditionally if you like. I was being a bit lazy perhaps?

PR #58, opened 2024-11-12
bump vite to `6.0.0-beta.9`

dario-piotrowicz:


packages/vite-plugin-cloudflare/src/runner/index.ts
```diff
@@ -161,6 +161,9 @@ export function createWorkerEntrypointWrapper(
 					server.accept();
 					createModuleRunner(this.env, server);
 
+					// TODO: here we need to add some validation, if createModuleRunner throws we want to return an error response
+					//       and process.exit or something from the other side
+
```
dario-piotrowicz, line 166: This is something I noticed when working on this PR, but I don't want to scope creep my PR here, I'll create a separate issue for this validation (which is actually unrelated to this PR anyways)

dario-piotrowicz: Issue created: https://github.com/flarelabs-net/vite-plugin-cloudflare/issues/59

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -204,41 +204,62 @@ export function getMiniflareOptions(
...
+							`unexpected payload event sent to the vite-invoke binding: "${payload.event}"`,
+						);
+
+						const invokePayloadData = payload.data as {
+							id: string;
+							name: string;
+							data: [
+								string,
+								string,
+								FetchFunctionOptions, // TODO: this should be the invoke equivalent
```
dario-piotrowicz, line 226: `FetchFunctionOptions` seems to be the correct type here actually! ([source](https://github.com/vitejs/vite/blob/584a573e1fdfa64018b2c56d147f2b4fec2d9733/packages/vite/src/shared/invokeMethods.ts#L83))

jamesopstad (commented):
The move to using a single transport for fetching modules and HMR in `6.0.0-beta.9` is quite a big change. It's worth reading through the PR and code changes (https://github.com/vitejs/vite/pull/18362).

What you've done here has inadvertently removed all support for HMR. I can't advise on what the best approach is overall until I've had a play around with it myself as there are a few possibilities.

I suggest we hold off on this for now and focus on merging the other PRs. It also reminded me of another longstanding issue we have with HMR so I've opened an issue for that (https://github.com/flarelabs-net/vite-plugin-cloudflare/issues/64).

dario-piotrowicz:
> What you've done here has inadvertently removed all support for HMR. I can't advise on what the best approach is overall until I've had a play around with it myself as there are a few possibilities.

Thanks for noticing that! Sorry I missed to check that as I didn't think any of my changes should effect HMR! (totally my bad 😓)

I will look into fixing HMR here as soon as I can.

By the way, I will also create an issue for HMR testing, I think we should definitely have some tests that do make sure that HMR works

dario-piotrowicz:
Closing in favour of #70

PR #62, opened 2024-11-12
Initial Assets support

### LINKED ISSUE #36: Add initial support for Workers Assets

jamesopstad:
resolves #36

This PR adds initial support for Workers Assets by covering the common use case of static sites (including SPAs) with and without user Workers.

The following features are not supported and will be added later:

- Client builds (this will be added in a follow-up PR along with the work for #61)
- Using the assets binding to load non-HTML assets (#60 requires discussion)


packages/vite-plugin-cloudflare/src/assets/asset-worker.ts
```diff
@@ -0,0 +1,44 @@
+// @ts-ignore
+import AssetWorker from '@cloudflare/workers-shared/dist/asset-worker.mjs';
```
jamesopstad, line 2: The workers from `@cloudflare/workers-shared` are not currently properly exported with types. When they are, they types in this file will be improved.

playground/static-mpa/__tests__/assets.spec.ts
jamesopstad, line 1: Just a note to say that at present tests like these can be dependant on order because the navigations are preserved. Should we update the testing configuration to call `page.goto(viteTestUrl)` before each test?

dario-piotrowicz: I also agree with this we should probably reset the page location per-test.

My only concern is that this can make the `serverLogs` a bit less stable since more `page.goto` calls can trigger more server logs. but that's not a huge issue and something we can deal with later when/if needed (I am also not 100% completely happy with `serverLogs` to be completely honest... but given the testing structure I am not sure if something better can be implemented)

packages/vite-plugin-cloudflare/src/assets/asset-worker.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export default class CustomAssetWorker extends (AssetWorker as typeof WorkerEntrypoint<Env>) {
+	override async fetch(request: Request): Promise<Response> {
+		const response = await super.fetch!(request);
+		const headers = new Headers(response.headers);
+		headers.delete('ETag');
+		headers.delete('Cache-Control');
+
+		// Do we need to do anything else here to duplicate the original response?
```
petebacondarwin, line 18: This appears to be the official approach to doing this:

https://developers.cloudflare.com/workers/examples/alter-headers/

packages/vite-plugin-cloudflare/src/assets/asset-worker.ts
```diff
@@ -0,0 +1,44 @@
...
+		});
+	}
+	async unstable_getByETag(
+		eTag: string,
+	): Promise<{ readableStream: ReadableStream; contentType: string }> {
+		const url = new URL(eTag, UNKNOWN_HOST);
+		const response = await this.env.__VITE_FETCH_ASSET__.fetch(url);
+
+		if (!response.body) {
+			throw new Error(`Unexpected error. No HTML found for ${eTag}.`);
```
petebacondarwin, line 31: Perhaps return a 500 response rather than throwing?

packages/vite-plugin-cloudflare/src/assets/asset-worker.ts
```diff
@@ -0,0 +1,44 @@
...
+		}
+
+		return { readableStream: response.body, contentType: 'text/html' };
+	}
+	async unstable_exists(pathname: string): Promise<string | null> {
+		// We need this regex to avoid getting `//` as a pathname, which results in an invalid URL. Should this be fixed upstream?
+		const url = new URL(pathname.replace(/^\/{2,}/, '/'), UNKNOWN_HOST);
```
petebacondarwin, line 38: Are you saying that the pathname passed to `unstable_exists()` sometimes contains `//`? Is it only the case when the path is "empty", i.e. `"//"`?
Definitely feels like a bug in the Asset Worker.

packages/vite-plugin-cloudflare/src/assets/asset-worker.ts
```diff
@@ -0,0 +1,44 @@
...
+		const response = await this.env.__VITE_ASSET_EXISTS__.fetch(url);
+		const exists = await response.json();
+
+		return exists ? pathname : null;
```
petebacondarwin, line 42: I assume that `exists` is always a boolean? Not just some random thing that could accidentally be falsy (like `""` or 0)?

packages/vite-plugin-cloudflare/src/assets/router-worker.ts
```diff
@@ -0,0 +1,4 @@
+// @ts-ignore
+import routerWorker from '@cloudflare/workers-shared/dist/router-worker.mjs';
+
+export default routerWorker;
```
petebacondarwin, line 4: This file is just to make the imports elsewhere in the code look cleaner?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -103,6 +105,7 @@ export function cloudflare<T extends Record<string, WorkerOptions>>(
 				getMiniflareOptions(normalizedPluginConfig, viteConfig, viteDevServer),
 			);
 
+			let routerWorker = await getRouterWorker(miniflare);
```
petebacondarwin, line 108: Is caching this object important?
Could we just call `getRouterWorker()` on each request inside the middleware?

packages/vite-plugin-cloudflare/src/assets.ts
```diff
@@ -0,0 +1,18 @@
...
+		| 'none';
+	notFoundHandling?: 'single-page-application' | '404-page' | 'none';
+}
+
+export const ROUTER_WORKER_NAME = '__router-worker__';
+export const ASSET_WORKER_NAME = '__asset-worker__';
+export const ASSET_WORKERS_COMPATIBILITY_DATE = '2024-10-04';
+
+export function getRouterWorker(miniflare: Miniflare) {
+	return miniflare.getWorker(ROUTER_WORKER_NAME);
```
petebacondarwin, line 17: So this implementation does not support multiple user workers having assets.
I think this is actually a real requirement going forward, even if we don't land it in this iteration.
I can be onboard with the idea that in most (all?) scenarios there is only one entry-point Worker that hosts assets (from the point of view of the Vite HTTP server).

But we will need to support servicing binding to Workers with Assets from the entry-point Worker.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -118,7 +126,95 @@ export function getMiniflareOptions(
 	viteConfig: vite.ResolvedConfig,
 	viteDevServer: vite.ViteDevServer,
 ): MiniflareOptions {
-	const workers = Object.values(normalizedPluginConfig.workers).map(
+	const assetWorkers: Array<WorkerOptions> = [
```
petebacondarwin, line 129: Should this stuff not be configurable?
I.e. not adding this stuff for Workers that do not have assets configured?

If I am building a Worker that does not have assets then I would not expect `vite dev` to serve assets from my local file system ahead of my Worker code.

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -1,11 +1,14 @@
 import * as path from 'node:path';
 import * as vite from 'vite';
 import { unstable_getMiniflareWorkerOptions } from 'wrangler';
+import type { AssetConfig } from './assets';
 import type { SourcelessWorkerOptions } from 'wrangler';
 
 export interface WorkerOptions {
 	main: string;
 	wranglerConfig?: string;
+	// TODO: tighten up types so assets can only be bound to entry worker
```
petebacondarwin, line 10: I'm not sure this should be a goal. Assets should be configurable per Worker.

packages/vite-plugin-cloudflare/src/assets/asset-worker.ts
jamesopstad: Thanks. I'll change it to use that approach.

packages/vite-plugin-cloudflare/src/assets/asset-worker.ts
jamesopstad: We can't really return a response here because the return type is `Promise<{ readableStream: ReadableStream; contentType: string }>`. The error should only occur if there's a regression in the code so would hopefully be caught in CI rather than by end users.

packages/vite-plugin-cloudflare/src/assets/asset-worker.ts
jamesopstad: Yes, so if the `pathname` of the request is `/` then one of the pathnames for the file that it will check is `//`. I haven't dug into exactly why this is.

packages/vite-plugin-cloudflare/src/assets/asset-worker.ts
jamesopstad: Yep it's a boolean:

https://github.com/flarelabs-net/vite-plugin-cloudflare/blob/ca5b0697b2914941e94483880b3f7c8775fc09e1/packages/vite-plugin-cloudflare/src/miniflare-options.ts#L188

packages/vite-plugin-cloudflare/src/assets/router-worker.ts
jamesopstad: It's so that all the internal workers used by the plugin are built and imported in a consistent way. So this file is the entry for the router worker build:

https://github.com/flarelabs-net/vite-plugin-cloudflare/blob/ca5b0697b2914941e94483880b3f7c8775fc09e1/packages/vite-plugin-cloudflare/tsup.config.ts#L22

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: I'm not sure how much of an advantage there is over doing this on every request. This would also mean calling `createMiddleware` on each request, which seems a bit wasteful. We can do that if you prefer though.

packages/vite-plugin-cloudflare/src/assets.ts
jamesopstad: This PR is very much about initial assets support to cover the most common use cases and meet the expectations of people used to using Vite.

Not ruling out multiple workers having assets at all but there are quite a few questions to answer before we do that and it will need its own dedicated PR.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: > Should this stuff not be configurable? I.e. not adding this stuff for Workers that do not have assets configured?

My instinct is that an assets only project (static sites etc.) should be zero config by default. If we needed to make it configurable, I would therefore argue that it should opt-out rather than opt-in. I'm not yet convinced that it's actually necessary though and think that we can consider that with broader questions about what the API should look like.

> If I am building a Worker that does not have assets then I would not expect `vite dev` to serve assets from my local file system ahead of my Worker code.

I think that's where we differ because that would always be my expectation about how Vite would work. This already happens before this PR and will continue to happen unless we remove the default middlewares. We may want to go with that approach in the future in order to allow the Worker to sit in front of all assets and for the plugin to take control all asset handling. We would need to look at the implications carefully though, particularly in relation to full-stack frameworks.

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: I don't disagree that the longer term goal should be for assets to be configurable per worker but I think that, even then, the entry worker may still be distinct in the way it needs to be configured. Again, that needs to be part of broader discussions about the API and I feel it's outside the scope of this PR.

packages/vite-plugin-cloudflare/src/assets/asset-worker.ts
jamesopstad: 08eb35e

playground/static-mpa/__tests__/assets.spec.ts
jamesopstad: I've made the change here - 336bc26.

Having looked at the test setup a little, I do wonder if we would be better off just using Playwright directly rather than this combination of Playwright and Vitest for the E2E tests. Not a priority now but something we might want to have a look at in the future as it could simplify things.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
petebacondarwin: So will `vite build` always output a Worker with Assets turned on, to account for the potential assets that are served by default in `vite dev`?

packages/vite-plugin-cloudflare/src/assets/asset-worker.ts
petebacondarwin: Fair enough, I see that we are throwing in production too: https://github.com/cloudflare/workers-sdk/blob/c1022d66c6536c9e8199ebffe73d2358aa4064ee/packages/workers-shared/asset-worker/src/index.ts#L149-L154

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Good point. That was my mistake. I'll change to go with this approach then (I presume it's very fast for Miniflare to get the Worker stub).

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: For standard Vite, the things that are included in the client build are anything that's imported by the entry file (`index.html` by default) and/or its sub-dependencies (including static assets) and the contents of the `public` directory (the name of this directory is configurable). This PR doesn't implement the build side of this but my thinking was that the build would check if the client entry file exists and, if it does, build the client environment and reference its output as the assets directory.

We need to have a chat about using the assets binding to load non-HTML assets and ensuring that these are included in the build. That will be a subsequent PR.

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: f2b1fd9

PR #68, opened 2024-11-19
Add support for `vite preview`

### LINKED ISSUE #61: Add support for `vite preview`

jamesopstad:
resolves #61

This adds support for `vite preview` as well as test coverage for build output. The details of the implementation are likely to change quite a bit when we have finalised the API and the structure of the build output and for the time being quite a few assumptions are made. For now, I have avoided outputting any configuration file and instead read from the existing `wrangler.toml` and Vite config.

The only build tests that do not currently pass are the 'module-resolution' tests and the 'node-compat/worker-basic' tests so these are currently omitted.


packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -148,7 +150,9 @@ export function createCloudflareEnvironmentOptions(
 					//       so the input value here serves both as the build input as well as the starting point for
 					//       dev pre-bundling crawling (were we not to set this input field we'd have to appropriately set
 					//       optimizeDeps.entries in the dev config)
-					input: options.main,
+					input: userConfig.root
+						? path.resolve(userConfig.root, options.main)
+						: path.resolve(options.main),
```
jamesopstad, line 155: It turns out that `rollupOptions.input` isn't automatically resolved relative to the root so I had to add the resolution here.

playground/module-resolution/__tests__/no-prebundling/module-respolution-no-prebundling.spec.ts
```diff
@@ -6,6 +6,7 @@ import {
 	viteTestUrl,
 } from '../../../__test-utils__';
 
+// TODO: test build
```
jamesopstad, line 9: This doesn't currently pass when the build output is tested. Perhaps @dario-piotrowicz  can look into this in a subsequent PR.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -364,6 +362,116 @@ export function getMiniflareOptions(
...
+						...assets,
+						name: 'assets-only',
+						modules: [
+							{
+								type: 'ESModule',
+								path: '__dummy-module__',
+								contents: '',
+							} as const,
+						],
+					},
```
dario-piotrowicz, line 434: I think something's wrong here 😕 

if I am understanding things correctly this is for assets-only projects right?

my issue is that if I comment this entire object out the build tests still pass 😕 

the `react-spa` ones should fail though, shouldn't they?

dario-piotrowicz: (in either case I think that all the code we add to the package, if removed should cause at least one test to fail, otherwise something's not right in my opinion (we either don't cover features or like now something is just off))

playground/module-resolution/__tests__/module-resolution.spec.ts
```diff
@@ -8,6 +8,7 @@ import {
 	viteTestUrl,
 } from '../../__test-utils__';
 
+// TODO: test build
```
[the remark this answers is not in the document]
dario-piotrowicz: sure! sounds good to me 🙂👍

playground/module-resolution/__tests__/no-prebundling/module-respolution-no-prebundling.spec.ts
dario-piotrowicz: sure! sounds good to me 🙂👍

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -364,6 +362,116 @@ export function getMiniflareOptions(
...
+		? normalizedPluginConfig.workers[normalizedPluginConfig.entryWorkerName]
+		: undefined;
+
+	const assetsDirectory = path.resolve(
+		viteConfig.root,
+		viteConfig.build.outDir,
+		'client',
+	);
+	const hasAssets = fs.existsSync(assetsDirectory);
+	const assets = hasAssets
```
dario-piotrowicz, line 379: Nit: I think a name like `assetsOptions` or something like that would be clearer here

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -364,6 +362,116 @@ export function getMiniflareOptions(
...
+					{
+						...assets,
+						name: 'assets-only',
+						modules: [
+							{
+								type: 'ESModule',
+								path: '__dummy-module__',
+								contents: '',
+							} as const,
+						],
```
dario-piotrowicz, line 433: I think this can be simplified
```suggestion
						script: '',
						modules: true,
```

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: I just tried commenting this object out and the react-spa and static-mpa tests do fail. What are the steps to recreate this? Are you sure you rebuilt the plugin after the change?

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -142,13 +144,17 @@ export function createCloudflareEnvironmentOptions(
 				createEnvironment(name, config) {
 					return new vite.BuildEnvironment(name, config);
 				},
+				// This is a bit of a hack to make sure the user can't override the output directory at the environment level
```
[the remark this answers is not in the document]
jamesopstad: Sorry, I haven't provided enough context here. I need to dig into exactly what the issue is but with the way we had things before, `environment.build.outDir` was getting overridden by `undefined` in preview mode. I've just worked around it for the moment but will have another look.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: b4f921a

packages/vite-plugin-cloudflare/src/miniflare-options.ts
dario-piotrowicz: no I just re-tried and it does seem to trigger some failures... I was quite sure that it didn't 😕  (I was building the plugin in watch mode, I wonder if that could have caused my confusion)

sorry, it's all good, my bad! 🙇

dario-piotrowicz (approved):
Looks great to me! it's very nice to have preview enabled! 🚀 🥳

playground/module-resolution/__tests__/module-resolution.spec.ts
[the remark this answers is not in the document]
jamesopstad: Added issue #71.

playground/node-compat/__tests__/worker-basic/basic.spec.ts
```diff
@@ -1,6 +1,7 @@
 import { expect, test } from 'vitest';
 import { getTextResponse, isBuild } from '../../../__test-utils__';
 
+// TODO: test build
```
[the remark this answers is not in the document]
jamesopstad: Added issue #72.

Issue #64, opened 2024-11-13
HMR implementation limited by restrictions to persistent WebSockets


jamesopstad:
This is something that I mentioned early on in this project but I'm creating this issue now to track it.

Currently, we are creating the module runner inside a Worker, despite the fact that the module runner is stateful and Workers are required to be stateless. As our HMR implementation requires a persistent WebSocket connection this leads to the following error in some scenarios (e.g. when using `import.meta.hot.send`):

_Cannot perform I/O on behalf of a different request. I/O objects (such as streams, request/response bodies, and others) created in the context of one request handler cannot be accessed from a different request's handler. This is a limitation of Cloudflare Workers which allows us to improve overall performance. (I/O type: Native)_

Others have tried to circumvent this by running user Worker code inside a Durable Object rather than a Worker. I don't think this is the right solution, however, as this would also allow user code to do things that would break in production. I think there may be a better solution where we still run the user code inside a Worker but create the module runner inside a Durable Object. I need to investigate further.

PR #69, opened 2024-11-20
Attempt at resolving #64

jamesopstad:
Hi @petebacondarwin and @dario-piotrowicz. I had a go at resolving #64 but haven't had any luck unfortunately. I'm sharing what I tried here to see if you have any suggestions. It seems that even if the WebSocket connection is created in a Durable Object then the same restrictions still apply when trying to use it from a Worker. I don't think running the user code inside a DO is an option because then they would also be able to evade this restriction and dev wouldn't accurately reflect production.


jamesopstad:
I've updated the description in #64 as I feel this may turn out to be a bit of a non-issue and not something we should worry about too much unless the use-case arises.

PR #84, opened 2024-12-03
New config API, `wrangler.json` output and updated preview mode

### LINKED ISSUE #72: Add build tests for node-compat/worker-basic playground
These tests pass in dev but not when the build output is tested. Needs investigating.

### LINKED ISSUE #77: Output `wrangler.json` as part of build process

jamesopstad:
resolves #77
resolves #72

This PR makes the following changes:
- The config now uses a new API that reads from the user's Wrangler config file(s)
- `wrangler.json` files are now generated and included as part of the build
- Preview mode has been updated to read from the generated `wrangler.json` file(s)

Notes:
- This currently uses a prerelease version of Wrangler that is created from this PR - https://github.com/cloudflare/workers-sdk/pull/7256
- A user Wrangler config file is currently required for all projects and it must include either an `assets` field or `name` and `main` fields. We may wish to add support for starting without a Wrangler config file for assets only projects.
- We currently enforce that the output for all packages is inside the same build directory. We may wish to allow more flexibility here but this requires being able to read environment specific build options in preview mode.
- Additional validation of the input Wrangler config (e.g. to warn on redundant fields) should be added in a follow up PR.
- Once this PR is merged, I'll do a follow up to remove the `invariant` function and use `assert` instead.


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -72,81 +94,135 @@ export function cloudflare<T extends Record<string, WorkerOptions>>(
...
+			if (resolvedPluginConfig.type === 'workers') {
+				const workerConfig =
+					resolvedPluginConfig.workers[this.environment.name];
+
+				if (!workerConfig) {
+					return;
 				}
+
+				// Currently assuming that the entrypoint of the build is `index.js`. Should we determine this from the config?
+				workerConfig.main = 'index.js';
```
petebacondarwin, line 161: I assume that this is configurable by the user?
Or maybe by other plugins?

So I would expect we should be able to grab this from the config somewhere...

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -179,19 +255,16 @@ export function cloudflare<T extends Record<string, WorkerOptions>>(
...
+				getPreviewMiniflareOptions(resolvedPluginConfig, vitePreviewServer),
 			);
 
-			const middleware = createMiddleware(
-				({ request }) => {
-					return miniflare.dispatchFetch(toMiniflareRequest(request), {
-						redirect: 'manual',
-					}) as any;
-				},
-				{ alwaysCallNext: false },
```
petebacondarwin, line 193: Out of curiosity - why is this no longer needed?

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -24,11 +25,16 @@ type PersistOptions = Pick<
 	| 'r2Persist'
 >;
 
-export function getPersistence(persistPath: string | false): PersistOptions {
-	if (persistPath === false) {
+export function getPersistence(
+	root: string,
+	persistState = true,
```
petebacondarwin, line 30: Do we ever call this with `persistState` set to false?

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Yeah, I had a quick look but it wasn't obvious to me where to find this. Will look into it more.

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: I think this is only relevant when you have other plugins that add middleware so I thought it's best to remove it for now and revisit when we look at framework integrations.

playground/node-compat/worker-crypto/wrangler.toml
```diff
@@ -1,2 +1,4 @@
+name = "worker"
```
dario-piotrowicz, line 1:
```suggestion
name = "worker-crypto"
```

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: Not at the moment. I would be inclined to add `persistState` as an option to the plugin config so left the option open. I think people didn't feel this was necessary when we discussed it the other day. Happy to go with what you prefer here.

No need for this function to be exported so I'll remove that.

playground/static-mpa/vite.config.ts
```diff
@@ -12,19 +12,11 @@ export default defineConfig({
 						contact: path.resolve(__dirname, 'contact.html'),
 						'404': path.resolve(__dirname, '404.html'),
 						about: path.resolve(__dirname, 'about/index.html'),
-						about404: path.resolve(__dirname, 'about/404.html'),
+						'about-404': path.resolve(__dirname, 'about/404.html'),
```
dario-piotrowicz, line 15: why this change?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -72,81 +94,135 @@ export function cloudflare<T extends Record<string, WorkerOptions>>(
...
-			viteConfig = resolvedConfig;
-			normalizedPluginConfig = normalizePluginConfig(
-				pluginConfig,
-				resolvedConfig,
-			);
+			if (resolvedPluginConfig.type === 'workers') {
+				options.build = {
+					...options.build,
+					// Puts all environment builds in subdirectories of the same build directory
+					// TODO: allow the user to override this
```
dario-piotrowicz, line 104: shall we create a gh issue?

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -1,101 +1,189 @@
...
+	userConfig: vite.UserConfig,
+	wranglerConfigPaths: Set<string>,
+	isEntryWorker?: boolean,
+): AssetsOnlyResult | WorkerResult {
+	if (wranglerConfigPaths.has(configPath)) {
+		throw new Error(`Duplicate Wrangler config path found: ${configPath}`);
+	}
+
+	const wranglerConfig = readConfig(configPath, {
+		// We use the mode from the user config rather than the resolved config for now so that the mode has to be set explicitly. Otherwise, some things don't work as expected when `development` and `production` environments are not present.
```
dario-piotrowicz, line 65: why is this? 😕

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -1,101 +1,189 @@
...
+	for (const extension of ['json', 'jsonc', 'toml']) {
+		const configPath = path.join(root, `wrangler.${extension}`);
+
+		if (fs.existsSync(configPath)) {
+			return configPath;
+		}
+	}
+}
+
+// Worker names can only contain alphanumeric characters and '_' whereas environment names can only contain alphanumeric characters and '$', '_'
```
dario-piotrowicz, line 114: do we know why environment names don't allow `-`? maybe that's something that we can change upstream? (it is pretty minor in any case 😅)

dario-piotrowicz, line 114:
```suggestion
// Worker names can only contain alphanumeric characters and '-' whereas environment names can only contain alphanumeric characters and '$', '_'
```

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -1,101 +1,189 @@
...
+function workerNameToEnvironmentName(workerName: string) {
+	return workerName.split('-').join('_');
```
dario-piotrowicz, line 116: nit
```suggestion
	return workerName.replaceAll('-', '_');
```

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -1,101 +1,189 @@
...
-			);
+	const root = userConfig.root ? path.resolve(userConfig.root) : process.cwd();
+
+	const configPath = pluginConfig.wranglerConfig
+		? path.join(root, pluginConfig.wranglerConfig)
+		: findWranglerConfig(root);
+
+	invariant(
+		configPath,
+		`Config not found. Have you created a wrangler.json or wrangler.toml file?`,
```
dario-piotrowicz, line 132:
```suggestion
		`Config not found. Have you created a wrangler.json(c) or wrangler.toml file?`,
```

dario-piotrowicz (commented):
Looks good to me 😄 

I just left a minor few comments 🙂 

By the way, I am not a big fan on relying on a wrangler prerelease... that feels super brittle to me.... but this PR can be quite a blocker for other work, so I think it'll be ok this time 😕  (but I think we should not merge this PR until https://github.com/cloudflare/workers-sdk/pull/7256 is at least approved)

jamesopstad:
> By the way, I am not a big fan on relying on a wrangler prerelease... that feels super brittle to me.... but this PR can be quite a blocker for other work, so I think it'll be ok this time 😕 (but I think we should not merge this PR until [cloudflare/workers-sdk#7256](https://github.com/cloudflare/workers-sdk/pull/7256) is at least approved)

I wouldn't say I'm a fan either but the two are intrinsically linked. The approach in this PR needs to be approved before the other PR can be moved forward and merged. I agree it would be best to merge the other PR before this one.

playground/module-resolution/vite.config.no-prebundling.ts
```diff
@@ -15,27 +15,17 @@ export default defineConfig({
...
-			persistTo: false,
-		}),
-	],
+			resolve: {
+				// external modules don't get prebundled
+				external: ['@cloudflare-dev-module-resolution/requires/ext'],
+			},
+		},
+	},
+	plugins: [cloudflare({ viteEnvironmentName: 'worker' })],
```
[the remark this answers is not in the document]
jamesopstad: Not sure what you mean by 'inline the env configs'. Can you explain?

I do think it might be good for us to allow setting some Vite environment options in the plugin config but that seems like something different.

playground/node-compat/worker-basic/wrangler.toml
```diff
@@ -1,2 +1,4 @@
+name = "worker"
```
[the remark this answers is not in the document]
jamesopstad: What's the reason for changing all these worker names? Does it matter?

playground/static-mpa/vite.config.ts
jamesopstad: Completely inconsequential. I just did it because I didn't like how it looked 😂.

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: #85

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: It's the way `readConfig` from Wrangler works, unfortunately. I was hoping that if an environment is specified and no environments are present in the config file then it would just use the top level config. This doesn't seem to be the case though.

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: I assume there's probably a reason. Not sure why though.

jamesopstad: 89e3ff4

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: 89e3ff4

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: 89e3ff4

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -179,19 +255,16 @@ export function cloudflare<T extends Record<string, WorkerOptions>>(
 				});
 			};
 		},
-		configurePreviewServer(vitePreviewServer) {
+		async configurePreviewServer(vitePreviewServer) {
```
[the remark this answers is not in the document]
jamesopstad: 825e6d5

playground/module-resolution/vite.config.no-prebundling.ts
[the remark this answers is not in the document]
dario-piotrowicz: with inline env config I mean setting the environment configs in the config option here, so instead of having:
```
environments: {
		worker: {
			<CONFIGS>
		},
},
plugins: [cloudflare({ viteEnvironmentName: 'worker' })], // <- the name is used to get the <CONFIGS>
```

it would be probably feel much more natural to me to have:
```
plugins: [cloudflare({ <CONFIGS> })], // (with an optional `viteEnvironmentName`)
```
or, to have a more clear distinction between configs have a specific field for this (probably better):
```
plugins: [cloudflare({ viteEnvironmentName: 'worker', environmentConfig: <CONFIGS> })],
```

playground/node-compat/worker-basic/wrangler.toml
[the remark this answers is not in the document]
dario-piotrowicz: mh... probably not... I mean... it can make things clearer if we were to ever deploy any of these... 🤷

it is a pretty minimal thing indeed... but also I don't see why not having more specific names 🤷

not a big deal, totally up to you, if you want to keep it to `worker` that's ok by me 🙂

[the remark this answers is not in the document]
jamesopstad: It's fine, I'm happy to do it. I just asked because the worker names haven't changed in this PR.

dario-piotrowicz (approved):
Looks good to me, approving but again I think that ideally we shouldn't merge this until https://github.com/cloudflare/workers-sdk/pull/7256 is merged (or close to)

packages/vite-plugin-cloudflare/src/plugin-config.ts
petebacondarwin: > I was hoping that if an environment is specified and no environments are present in the config file then it would just use the top level config

I am pretty sure that is how it works. It just warns but doesn't error: https://github.com/cloudflare/workers-sdk/blob/b5c72af7ea061cae9948ae51e3252d8857643890/packages/wrangler/src/config/validation.ts#L245

playground/node-compat/worker-basic/wrangler.toml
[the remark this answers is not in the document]
dario-piotrowicz: ah.... sorry I missed that 😅

well again, whatever you prefer I'll be fine with 🙂👍

packages/vite-plugin-cloudflare/src/plugin-config.ts
dario-piotrowicz: https://github.com/vitejs/vite/blob/81c31c16d10bff48c9965f7aedbf7efb7c7022c4/packages/vite/src/node/baseEnvironment.ts#L57-L58

feels like a pretty arbitrary reasoning... 🤷 

but hey, we can live with that 😅 🤷

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: If that was the case then you'd be able to the change `env` here from `userConfig.mode` to some arbitrary value and all the tests would still pass.

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: Hmm, I wouldn't have thought that `-` needs escaping in directory names. It would be nicer if they could be used so maybe it is worth looking into changing.

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: This comment explains the behaviour here:

https://github.com/cloudflare/workers-sdk/blob/bea65583f72fcf8346ed82d4b9b1176669bcfd23/packages/wrangler/src/config/validation.ts#L193-L205

For Workers, if an environment name is specified and the environment is not configured then a fake empty environment is used in its place. This has the effect of removing inheritable properties that are set at the top level. You can see the relevant code here (the third argument is the dummy env):

https://github.com/cloudflare/workers-sdk/blob/bea65583f72fcf8346ed82d4b9b1176669bcfd23/packages/wrangler/src/config/validation.ts#L219-L228

What we really want is the behaviour that is used for Pages but it feels like this might be a breaking change.

jamesopstad:
It turns out that this also resolves #72. The node-compat/worker-basic build test now passes. Not entirely sure why but I'll take it 😀.

PR #86, opened 2024-12-04
package.json improvements

jamesopstad:
Just some minor changes to make the package more ready to be used.

Are we happy with just publishing ESM? I would prefer that to avoid the extra complexity. I ran the package through 'Are the types wrong?' and that's the only issue that came up.

https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/CJSResolvesToESM.md

### DIFF
--- a/packages/typescript-config/package.json
+++ b/packages/typescript-config/package.json
@@ -1,4 +1,5 @@
 {
 	"name": "@vite-plugin-cloudflare/typescript-config",
+	"version": "0.0.0",
 	"private": true
 }
--- a/packages/vite-plugin-cloudflare/package.json
+++ b/packages/vite-plugin-cloudflare/package.json
@@ -1,6 +1,14 @@
 {
 	"name": "@flarelabs-net/vite-plugin-cloudflare",
 	"version": "0.0.0",
+	"description": "Cloudflare plugin for Vite",
+	"keywords": [
+		"cloudflare",
+		"workers",
+		"cloudflare-workers",
+		"vite",
+		"vite-plugin"
+	],
 	"repository": {
 		"type": "git",
 		"url": "https://github.com/flarelabs-net/vite-plugin-cloudflare",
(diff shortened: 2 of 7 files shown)


packages/typescript-config/package.json
```diff
@@ -1,4 +1,5 @@
 {
 	"name": "@vite-plugin-cloudflare/typescript-config",
+	"version": "0.0.0",
```
petebacondarwin, line 3: 🎉

petebacondarwin (approved):
> Are we happy with just publishing ESM? I would prefer that to avoid the extra complexity. I ran the package through 'Are the types wrong?' and that's the only issue that came up.

I am pretty happy to avoid CJS.

packages/vite-plugin-cloudflare/package.json
```diff
@@ -9,23 +17,32 @@
...
 	"sideEffects": false,
 	"type": "module",
-	"main": "dist/index.js",
-	"types": "dist/index.d.ts",
+	"exports": {
+		".": {
+			"types": "./dist/index.d.ts",
+			"import": "./dist/index.js"
+		}
+	},
```
dario-piotrowicz, line 25: out of curiosity, is `exports` needed when we have only the index files? (and they are already specified by `main` and `types`)

jamesopstad: Not strictly. If we were to choose though I would go with `exports` as it's the more modern/encouraged approach and makes it easier to add additional exports in future. I left `main` in because there's no harm in doing so (`exports` is prioritised) but I can remove it if you like.

dario-piotrowicz: ah no, no problem I was just curious 🙂 

(and I can't never remember which is the new/modern one and which is the old one 😓)

PR #92, opened 2024-12-06
Remove invariant function and use node:assert in its place

jamesopstad:


dario-piotrowicz (approved):

petebacondarwin (approved):
So happy about this 🫶

PR #94, opened 2024-12-06
replace `process.env.NODE_ENV` at build time and enable react module-resolution build test

### LINKED ISSUE #82: In the module resolution playground app the react import works during dev but not build
If you run the module resolution playground app in dev mode everything seems to be working fine

But if you run it in preview mode and fetch from `/third-party/react` you get the following error:
Screenshot 2024-12-02 at 10 19 31

(which is something we've seen in the past: https://github.com/flarelabs-net/vite-plugin-cloudflare/issues/22)

Something is not adding up here to me, because in the module resolution app I don't see us opting in nodejs_compat anywhere, so should this issue not be present for `dev` as well?

Additionally if I add
```
compatibility_flags = ['nodejs_compat']
```
to the `wrangler.toml` file, then both for dev and build the app breaks completely erroring that the `unenv/runtime/node/process/$cloudflare` module can't be found

dario-piotrowicz:
resolves #82

### DIFF
--- a/packages/vite-plugin-cloudflare/package.json
+++ b/packages/vite-plugin-cloudflare/package.json
@@ -36,6 +36,7 @@
 	},
 	"dependencies": {
 		"@hattip/adapter-node": "^0.0.49",
+		"@rollup/plugin-replace": "^6.0.1",
 		"miniflare": "^3.20241205.0"
 	},
 	"devDependencies": {
--- a/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
+++ b/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
@@ -1,5 +1,6 @@
 import assert from 'node:assert';
 import { builtinModules } from 'node:module';
+import replace from '@rollup/plugin-replace';
 import * as vite from 'vite';
 import { getNodeCompatExternals } from './node-js-compat';
 import { INIT_PATH, UNKNOWN_HOST } from './shared';
@@ -153,6 +154,13 @@ export function createCloudflareEnvironmentOptions(
 				//       optimizeDeps.entries in the dev config)
 				input: workerConfig.main,
 				external: [...cloudflareBuiltInModules, ...getNodeCompatExternals()],
+				plugins: [
+					replace({
+						'process.env.NODE_ENV': JSON.stringify(
+							process.env.NODE_ENV ?? 'production',
+						),
+					}),
+				],
 			},
 		},
 		optimizeDeps: {
--- a/playground/module-resolution/__tests__/module-resolution.spec.ts
+++ b/playground/module-resolution/__tests__/module-resolution.spec.ts
@@ -76,9 +76,7 @@ describe('module resolution', async () => {
 	 *  special meaning to us.
 	 */
 	describe('third party packages resolutions', () => {
-		// TODO: we skip this test on build because a `ReferenceError: process is not defined` is thrown
-		//       (https://github.com/flarelabs-net/vite-plugin-cloudflare/issues/82)
-		test.skipIf(isBuild)('react', async () => {
+		test('react', async () => {
 			const result = await getJsonResponse('/third-party/react');
 			expect(result).toEqual({
 				'(react) reactVersionsMatch': true,
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -68,6 +68,9 @@ importers:
       '@hattip/adapter-node':
         specifier: ^0.0.49
         version: 0.0.49
+      '@rollup/plugin-replace':
+        specifier: ^6.0.1
+        version: 6.0.1(rollup@4.24.0)
       miniflare:
         specifier: ^3.20241205.0
         version: 3.20241205.0
@@ -1335,6 +1338,24 @@ packages:
       typescript:
         optional: true
 
+  '@rollup/plugin-replace@6.0.1':
+    resolution: {integrity: sha512-2sPh9b73dj5IxuMmDAsQWVFT7mR+yoHweBaXG2W/R8vQ+IWZlnaI7BR7J6EguVQUp1hd8Z7XuozpDjEKQAAC2Q==}
+    engines: {node: '>=14.0.0'}
+    peerDependencies:
+      rollup: ^1.20.0||^2.0.0||^3.0.0||^4.0.0
+    peerDependenciesMeta:
+      rollup:
+        optional: true
+
+  '@rollup/pluginutils@5.1.3':
+    resolution: {integrity: sha512-Pnsb6f32CD2W3uCaLZIzDmeFyQ2b8UWMFI7xtwUezpcGBDVDW6y9XgAWIlARiGAo6eNF5FK5aQTr0LFyNyqq5A==}
+    engines: {node: '>=14.0.0'}
+    peerDependencies:
+      rollup: ^1.20.0||^2.0.0||^3.0.0||^4.0.0
+    peerDependenciesMeta:
+      rollup:
+        optional: true
+
   '@rollup/rollup-android-arm-eabi@4.22.4':
     resolution: {integrity: sha512-Fxamp4aEZnfPOcGA8KSNEohV8hX7zVHOemC8jVBoBUHu5zpJK/Eu3uJwt6BMgy9fkvzxDaurgj96F/NiLukF2w==}
     cpu: [arm]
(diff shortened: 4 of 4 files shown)


jamesopstad:
According to this comment, Vite replaces `process.env.NODE_ENV` when `keepProcessEnv` is `false`.

https://discord.com/channels/804011606160703521/1314204727621193788/1314404199747813446

I'm wondering if it would be better to only set `keepProcessEnv: true` when `nodejs_compat` is enabled.

dario-piotrowicz:
> According to this comment, Vite replaces `process.env.NODE_ENV` when `keepProcessEnv` is `false`.
> 
> https://discord.com/channels/804011606160703521/1314204727621193788/1314404199747813446
> 
> I'm wondering if it would be better to only set `keepProcessEnv: true` when `nodejs_compat` is enabled?

Sorry I am not sure I follow

Specifically thinking of the react use case, we want react to work both with and without `nodejs_compat`, but in either case it requires us to replace its `process.env.NODE_ENV`, right? so I think that either way we do need the replacement introduced in this PR? are you suggesting to add the conditional setting of `keepProcessEnv` on top of the current changes? in that case what would the gain be?
Or am I missing something? 🤔

jamesopstad:
Ah, I probably misunderstood. I thought this example worked if `nodejs_compat` was enabled. But I'm guessing you mean that `process.env.NODE_ENV` doesn't exist even if `nodejs_compat` is enabled, in which case you're right.

dario-piotrowicz:
> Ah, I probably misunderstood. I thought this example worked if `nodejs_compat` was enabled.

I don't actually know, I'm unable to see if this works with `nodejs_compat` since when I try to set the flag I get the following error:
```
Cannot find module 'unenv/runtime/node/process/$cloudflare' imported from '/Users/dario/Repos/vite-plugin-cloudflare/playground-temp/module-resolution/src/index.ts
```

> But I'm guessing you mean that `process.env.NODE_ENV` doesn't exist even if `nodejs_compat` is enabled, in which case you're right.

Ah, sorry I see what you meant now, you thought that `process.env.NODE_ENV` could be just present with `nodejs_compat`? I don't think it is:
Screenshot 2024-12-09 at 11 06 12

maybe it should? WDYT? @petebacondarwin?

jamesopstad:
I think you're approach is correct. Even if `process.env.NODE_ENV` was available we'd want it to be statically replaced so that people don't end up with development builds of libraries (e.g. React) in production builds of their workers.

I sort of wondered if we should actually always be using `keepProcessEnv: false` but I guess people might use it for other things that aren't available at build time.

dario-piotrowicz:
> I think you're approach is correct. Even if `process.env.NODE_ENV` was available we'd want it to be statically replaced so that people don't end up with development builds of libraries (e.g. React) in production builds of their workers.

oh yeah! treeshaking slipped my mind for a sec! 😓 :+1:

petebacondarwin (approved):

jamesopstad:
@petebacondarwin @dario-piotrowicz Please can you explain what would happen in our case if we always had `keepProcessEnv: false` and how that relates to this - https://developers.cloudflare.com/workers/runtime-apis/nodejs/process/#processenv.

dario-piotrowicz:
> @petebacondarwin @dario-piotrowicz Please can you explain what would happen in our case if we always had `keepProcessEnv: false` and how that relates to this - https://developers.cloudflare.com/workers/runtime-apis/nodejs/process/#processenv.

This is, as far as I can tell, the only place in which `keepProcessEnv` is currently used: [`...(keepProcessEnv ? {} : processEnv),`](https://github.com/vitejs/vite/blob/258cdd637d1ee80a3c4571685135e89fe283f3a6/packages/vite/src/node/plugins/define.ts#L64) (alongside the upcoming changes in https://github.com/vitejs/vite/pull/18899), so all it does is adding these `define`s: 
```ts
Object.assign(processEnv, {
  'process.env': `{}`,
  'global.process.env': `{}`,
  'globalThis.process.env': `{}`,
  'process.env.NODE_ENV': JSON.stringify(nodeEnv),
  'global.process.env.NODE_ENV': JSON.stringify(nodeEnv),
  'globalThis.process.env.NODE_ENV': JSON.stringify(nodeEnv),
})
```
([source](https://github.com/vitejs/vite/blob/258cdd637d1ee80a3c4571685135e89fe283f3a6/packages/vite/src/node/plugins/define.ts#L23-L30))

As per the [Vite docs](https://vite.dev/config/shared-options.html#define): `Define global constant replacements. Entries will be defined as globals during dev and statically replaced during build`

My understanding is that `keepProcess: false` would cause all the `define`s listed above to be replaced/controlled by Vite (with some global constants in dev and with static replacements in build) meaning that [workerd's `process.env`](https://developers.cloudflare.com/workers/runtime-apis/nodejs/process/#processenv) would be completely out of the picture, both for dev but especially in production.

So I think that basically with `keepProcess: false` would completely cause workerd's `process.env` to be overshadowed by vite's own `process.env` implementation

dario-piotrowicz:
> > @petebacondarwin @dario-piotrowicz Please can you explain what would happen in our case if we always had `keepProcessEnv: false` and how that relates to this - https://developers.cloudflare.com/workers/runtime-apis/nodejs/process/#processenv.
> 
> This is, as far as I can tell, the only place in which `keepProcessEnv` is currently used: [`...(keepProcessEnv ? {} : processEnv),`](https://github.com/vitejs/vite/blob/258cdd637d1ee80a3c4571685135e89fe283f3a6/packages/vite/src/node/plugins/define.ts#L64) (alongside the upcoming changes in [vitejs/vite#18899](https://github.com/vitejs/vite/pull/18899)), so all it does is adding these `define`s:
> 
> ```ts
> Object.assign(processEnv, {
>   'process.env': `{}`,
>   'global.process.env': `{}`,
>   'globalThis.process.env': `{}`,
>   'process.env.NODE_ENV': JSON.stringify(nodeEnv),
>   'global.process.env.NODE_ENV': JSON.stringify(nodeEnv),
>   'globalThis.process.env.NODE_ENV': JSON.stringify(nodeEnv),
> })
> ```
> 
> ([source](https://github.com/vitejs/vite/blob/258cdd637d1ee80a3c4571685135e89fe283f3a6/packages/vite/src/node/plugins/define.ts#L23-L30))
> 
> As per the [Vite docs](https://vite.dev/config/shared-options.html#define): `Define global constant replacements. Entries will be defined as globals during dev and statically replaced during build`
> 
> My understanding is that `keepProcess: false` would cause all the `define`s listed above to be replaced/controlled by Vite (with some global constants in dev and with static replacements in build) meaning that [workerd's `process.env`](https://developers.cloudflare.com/workers/runtime-apis/nodejs/process/#processenv) would be completely out of the picture, both for dev but especially in production.
> 
> So I think that basically with `keepProcess: false` would completely cause workerd's `process.env` to be overshadowed by vite's own `process.env` implementation

(👆 if you want I can spin up a new project and corroborate the above)

jamesopstad:
I did just try setting `keepProcessEnv: false` and running all the tests. They all passed (including `node-compat/worker-process`). So either something else is going on here or we don't have test coverage for this.

This also meant I could enable the React module-resolution test and it passes without any changes.

dario-piotrowicz:
> I did just try setting `keepProcessEnv: false` and running all the tests. They all passed (including `node-compat/worker-process`). So either something else is going on here or we don't have test coverage for this.

yeah I suspect that we don't have test coverage for this 👍

I'll have a quick look

dario-piotrowicz:
> > I did just try setting `keepProcessEnv: false` and running all the tests. They all passed (including `node-compat/worker-process`). So either something else is going on here or we don't have test coverage for this.
> 
> yeah I suspect that we don't have test coverage for this 👍
> 
> I'll have a quick look

I had a quick look and I noticed that I actually misread [the docs](https://developers.cloudflare.com/workers/runtime-apis/nodejs/process/#processenv) (sorry totally my bad) 😓 

Screenshot 2024-12-10 at 17 01 00

in workers `process.env` is an empty object, so I think it's actually ok to just use `keepProcessEnv: false` since Vite replaces existing `process.env` with empty objects, so there is actually no real difference right?

`keepProcessEnv: true` could be needed if the workerd `process.env` workerd implementation were to change at some point, but yeah right not `keepProcessEnv: false` might be appropriate 👍

dario-piotrowicz:
closing in favour of #107 (since the core of the PR changes I figured it would make sense to just open a new one, I hope you guys done mind 😅)

PR #99, opened 2024-12-09
Restart dev server on Worker config changes

### LINKED ISSUE #54: restart vite dev when compat flags change
If the users changes the compat flags to cause node compatibility to toggle on or off, then we need to restart the dev server.
See https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/49#discussion_r1830973725

jamesopstad:
resolves #54

This simplifies things quite a bit. I'm also cleaning up the Miniflare instance on restarts now.

Note that the removal of the `assets.directory` field in some of playgrounds isn't related to this PR (meant to do it in the last one).

### DIFF
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -25,7 +25,7 @@ import type { Unstable_RawConfig } from 'wrangler';
 
 export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin {
 	let resolvedPluginConfig: ResolvedPluginConfig;
-	let miniflare: Miniflare;
+	let miniflare: Miniflare | undefined;
 
 	return {
 		name: 'vite-plugin-cloudflare',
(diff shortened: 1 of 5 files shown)


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -198,54 +197,33 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin {
...
-			let error: unknown;
-
-			const miniflare = new Miniflare(
+			miniflare = new Miniflare(
 				getDevMiniflareOptions(resolvedPluginConfig, viteDevServer),
 			);
 
 			await initRunners(resolvedPluginConfig, viteDevServer, miniflare);
-
-			viteDevServer.watcher.on('all', async (_, path) => {
```
dario-piotrowicz, line 210: ah ok, so we were already restarting on wrangler config changes! 🤔

dario-piotrowicz (approved):
Nice 😄

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Yes, but we were restarting Miniflare but not Vite. Now that `name`, `main` etc. are read from the Wrangler config file, that's not enough.

dario-piotrowicz: ah yeah yeah I remember now 🙂👍 

(anyways it's nice to see that the functionality we want is all there but we get simpler code ❤️)

PR #102, opened 2024-12-10
fix: ensure that the unenv package is available at runtime

petebacondarwin:
By moving the unenv dep into the dependencies rather than devDependencies, it is installed at runtime for the user of the plugin.

Otherwise we get errors like:

```
 Internal server error: Cannot find module 'unenv/runtime/node/process/$cloudflare' imported from '~/src/index.ts'
```


dario-piotrowicz (approved):
looks good to me 🙂

petebacondarwin:
This does not solve the problem for pnpm projects. Closing and looking for a better solution.

Issue #63, opened 2024-11-13
Investigate alternatives to stripping `/@id/` prefixes when fetching modules


jamesopstad:
See https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/57/files/1b8725bf6a9234e0185495173dfd12f37c42fc07#r1840067164

jamesopstad:
This approach has been removed as it had unintended side-effects. The `nodejs_compat` test have been temporarily disabled and should be reenabled as part of this PR.

PR #104, opened 2024-12-10
Fix virtual modules not working

### LINKED ISSUE #101: Virtual modules are not supported
Virtual modules are currently not supported. I imagine the `nodejs_compat` implementation is blocking them. As part of the fix, we should add a `virtual-modules` playground so we have test coverage.

jamesopstad:
resolves #101

This removes the code that was causing virtual modules to fail and adds a `virtual-modules` playground. I have disabled the `nodejs_compat` tests for now. #63 needs resolving before they can be reenabled.


packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -303,14 +303,14 @@ export function getDevMiniflareOptions(
...
 									// that start with `/@id/...`.
-									if (moduleId.startsWith('/@id/')) {
-										const result = {
-											externalize: moduleId.slice('/@id/'.length),
-											type: 'builtin',
-										} satisfies vite.FetchResult;
-
-										return new MiniflareResponse(JSON.stringify({ result }));
-									}
+									// if (moduleId.startsWith('/@id/')) {
```
petebacondarwin, line 306: Can you add a comment here explaining why this was here and why it has been commented out?

playground/node-compat/__tests__/worker-basic/basic.spec.ts
```diff
@@ -1,7 +1,8 @@
 import { expect, test } from 'vitest';
 import { getTextResponse } from '../../../__test-utils__';
 
-test('basic nodejs properties', async () => {
+// TODO: reintroduce test
```
petebacondarwin, line 4: Link to the issue?

playground/node-compat/__tests__/worker-random/random.spec.ts
```diff
@@ -3,15 +3,18 @@ import { getJsonResponse } from '../../../__test-utils__';
...
-			expect.any(String),
-			expect.any(String),
-			expect.any(String),
-			expect.any(String),
-		]);
-	},
-);
+// TODO: reintroduce test
+test
+	.runIf(!process.env.CI)
```
petebacondarwin, line 8: Actually (not related to this PR) this condition is not relevant to this test. It was copied incorrectly from the postgres test. We should remove this and also the comment above.

petebacondarwin (approved):
A few NITs but generally happy to go with this.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: c16f92f

playground/node-compat/__tests__/worker-basic/basic.spec.ts
jamesopstad: c16f92f

PR #105, opened 2024-12-10
Set persistState to false in playgrounds

jamesopstad:
I noticed that if the tests reran then stateful tests (e.g. durable objects) would fail as the state was being preserved. This sets `persistState: false` in all playgrounds and removes an error in the plugin TypeScript that meant `persistState` was a config option in auxiliary workers.


dario-piotrowicz (approved):
LGTM 🙂

petebacondarwin (approved):

PR #107, opened 2024-12-10
set `keepProcessEnv` to `false` and re-enable `react` module-resolution test

### LINKED ISSUE #82: In the module resolution playground app the react import works during dev but not build

dario-piotrowicz:
resolves #82

___

As discussed in #94 

As per [our docs](https://developers.cloudflare.com/workers/runtime-apis/nodejs/process/#processenv):
Screenshot 2024-12-10 at 17 01 00

in workers `process.env` is an empty object, so I think it's actually ok to just use `keepProcessEnv: false` since [Vite replaces `process.env`s with empty objects](https://github.com/vitejs/vite/blob/258cdd637d1ee80a3c4571685135e89fe283f3a6/packages/vite/src/node/plugins/define.ts#L23-L30), so I think that there is actually little to no difference between having `keepProcessEnv` set to `true` or `false` for us

Setting if to `false` has the benefit that Vite replaces `process.env.NODE_ENV` when `keepProcessEnv` is `false` ([discord message](
https://discord.com/channels/804011606160703521/1314204727621193788/1314404199747813446)) fixing the issue we were encountering with the `react` import

`keepProcessEnv: true` could be needed if the workerd `process.env` implementation were to change at some point, but right now `keepProcessEnv: false` might be the most appropriate configuration

### DIFF
--- a/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
+++ b/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
@@ -179,7 +179,6 @@ export function createCloudflareEnvironmentOptions(
 				],
 			},
 		},
-		keepProcessEnv: false,
 	};
 }
 
--- a/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
+++ b/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
@@ -179,7 +179,7 @@ export function createCloudflareEnvironmentOptions(
 				],
 			},
 		},
-		keepProcessEnv: true,
+		keepProcessEnv: false,
 	};
 }
 
--- a/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
+++ b/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
@@ -179,6 +179,7 @@ export function createCloudflareEnvironmentOptions(
 				],
 			},
 		},
+		keepProcessEnv: false,
 	};
 }
 
--- a/playground/module-resolution/__tests__/module-resolution.spec.ts
+++ b/playground/module-resolution/__tests__/module-resolution.spec.ts
@@ -76,9 +76,7 @@ describe('module resolution', async () => {
 	 *  special meaning to us.
 	 */
 	describe('third party packages resolutions', () => {
-		// TODO: we skip this test on build because a `ReferenceError: process is not defined` is thrown
-		//       (https://github.com/flarelabs-net/vite-plugin-cloudflare/issues/82)
-		test.skipIf(isBuild)('react', async () => {
+		test('react', async () => {
 			const result = await getJsonResponse('/third-party/react');
 			expect(result).toEqual({
 				'(react) reactVersionsMatch': true,


jamesopstad (approved):
Looks good to me. Do you mind adding some context to the description based on what we discussed in the other PR (and the Cloudflare docs)? Then @petebacondarwin can review and understand the implications. Thanks.

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -179,7 +179,7 @@ export function createCloudflareEnvironmentOptions(
 				],
 			},
 		},
-		keepProcessEnv: true,
+		keepProcessEnv: false,
```
petebacondarwin, line 182: Do we want to hard code this to false or just not set it at all?
What is the default for this value?
What would happen if the user specifies their own value for this?

petebacondarwin (approved):
If it doesn't fail tests and seems to work. Then this seems like a reasonable change to make.

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
dario-piotrowicz: > Do we want to hard code this to false or just not set it at all? What is the default for this value?

Ah yeah... very good point! 👍  `false` (or more precisely `undefined`/falsy: https://github.com/vitejs/vite/blob/63b82f1e29a00d06a82144fd03ea8d6eff114290/packages/vite/src/node/config.ts#L255-L259) is the default... so we might just remove this altogether...

Setting it to `false` makes it very explicit that we're relying on this behavior... but it does feel unnecessary... I don't have a preference here, we can either keep or remove it, what's your preference? 🙂 

 > What would happen if the user specifies their own value for this?

We don't have the `overrides` field anymore so I am not sure if the user can override this 🤔 (I'd imagine that @jamesopstad experimented with this so he should know).
But even if they were to override it not much would change I think, It would likely break the react code but that's about it. The exact same things (user overrides potentially breaking functionalities) would apply to all the other options we set here by the way.

jamesopstad: I'm happy with removing it altogether.

Need to have a proper look at how we're providing defaults and if they can be overridden as needed but can do that once this is merged.

dario-piotrowicz: removed 🙂

dario-piotrowicz: no my bad, it seems like it does need to be set to `false` 😕

PR #110, opened 2024-12-11
Allow output directories to be overridden

### LINKED ISSUE #85: Allow the user to set environment specific output directories
Currently, it is enforced that build output directories are subdirectories of the same build directory (specified by `build.outDir`). We need to allow more flexibility around this but this requires being able to read the environment specific output directories in preview mode.

jamesopstad:
resolves #85

This allows output directories to be overridden for each environment and corresponding tests. Somewhere along the way, the problems I was previously encountering have disappeared.

One thing to flag up is that if the user sets `build.outDir` at the root (i.e. not per environment) then multiple workers would go in the same directory and overwrite each other. I think this is OK and that it is expected that per environment output directories should be specified if using multiple environments.

### DIFF
--- a/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
+++ b/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
@@ -124,7 +124,6 @@ const cloudflareBuiltInModules = [
 
 export function createCloudflareEnvironmentOptions(
 	workerConfig: WorkerConfig,
-	userConfig: vite.UserConfig,
 ): vite.EnvironmentOptions {
 	return {
 		resolve: {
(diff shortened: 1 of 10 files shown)


petebacondarwin:
> One thing to flag up is that if the user sets build.outDir at the root (i.e. not per environment) then multiple workers would go in the same directory and overwrite each other. 

I don't suppose we could identify this case and then use the `build.outDir` at the root as a base for the automatic per-environment out directories?

dario-piotrowicz:
> One thing to flag up is that if the user sets build.outDir at the root (i.e. not per environment) then multiple workers would go in the same directory and overwrite each other. I think this is OK and that it is expected that per environment output directories should be specified if using multiple environments.

Shall we check if `build.outDir` is set at the root and in that case present a warning to the user?

jamesopstad:
> > One thing to flag up is that if the user sets build.outDir at the root (i.e. not per environment) then multiple workers would go in the same directory and overwrite each other.
> 
> I don't suppose we could identify this case and then use the `build.outDir` at the root as a base for the automatic per-environment out directories?

I couldn't find a straightforward way to do this because the per environment output directory you get in the `configEnvironment` options could come from either `build.outDir` or `environments[environment].build.outDir`. Can investigate more though.

petebacondarwin (approved):

jamesopstad:
Going to merge this and we can remove if it becomes the default behaviour in Vite.

PR #111, opened 2024-12-11
fix incorrect describe import

dario-piotrowicz:
Sorry my bad, I did not correct vscode's autoimport 😓


petebacondarwin (approved):
🎉 

I wonder if we could implement an eslint check that would prevent such imports?

dario-piotrowicz:
> 🎉
> 
> I wonder if we could implement an eslint check that would prevent such imports?

yeah I'd imagine so 🙂 

as discussed we don't have eslint set up yet though 😅

Issue #97, opened 2024-12-09
Remove unsupported Wrangler config options and in some cases warn when they are used


jamesopstad:
The following user defined Wrangler config fields are not applicable when using the Vite plugin:

alias, base_dir, build, define, dev, find_additional_modules, minify, no_bundle, node_compat (old), preserve_file_names, rules, site, tsconfig, upload_source_maps

Some are replaced by Vite functionality (e.g. alias, define etc.), some are not relevant (e.g. node_compat) and some are added to the config by the plugin at build time (e.g. no_bundle, rules).

We should define and implement the correct behaviour in each case.

PR #112, opened 2024-12-11
show warnings in the terminal when there are non-applicable configs in wrangler config files

dario-piotrowicz:
This PR adds warnings for when configs present in wrangler.toml/json(c) files are not applicable

Result:
Screenshot 2024-12-11 at 16 54 41

I've opened this PR to address #97, but I am not sure if it fully does because my implementation consists in simply ignoring all the non-applicable configs and show an appropriate warning in such case (which, in my opinion is an appropriate behavior?). Instead in #97 @jamesopstad seemed to suggest that we should have different handling for different non-applicable settings.

However I've structured the code in such a way that modifying it to add some custom behavior for non-applicable configs should be pretty simple (so even if this doesn't fully address #97 it should be a valid starting point for that).

What do you think @jamesopstad?


playground/multi-worker/__tests__/with-wrangler-warning/multi-worker.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { describe, expect, test } from 'vitest';
+import { getJsonResponse, isBuild, serverLogs } from '../../../__test-utils__';
+
+describe('multi-worker basic functionality', async () => {
+	test('wrangler warnings are present in the terminal', async () => {
```
jamesopstad, line 5: I think we should only run this test in dev. We shouldn't be validating or warning on the output configuration that we generate.

jamesopstad (changes requested):
Thanks Dario. I've added a few comments that I think need resolving.

I think this would benefit from some refactoring, renaming etc. and possibly using an approach more similar to the diagnostics in Wrangler. It's not a big deal now though and we can look at those in a follow up PR later.

playground/multi-worker/__tests__/with-wrangler-warning/multi-worker.spec.ts
dario-piotrowicz: yes, that's my bad, as I mentioned I didn't think of the preview config being generated by us when working on this 👍

packages/vite-plugin-cloudflare/src/miniflare-options.ts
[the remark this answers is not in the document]
dario-piotrowicz: No sorry, I replied to your comment too quickly, I think that the changes in this file are totally ok, as `readWorkerConfig` collects all the information (including the non-applicable configs) but doesn't do anything with them, it's up to the caller to the show warnings if needed (and perf wise collecting all the info is at most very neglibible).

So I don't see any issue in using `readWorkerConfig` here, on the contrary, I did set this in this way thinking that we should not use `unstable_readConfig` directly anymore but only use it thorough this `readWorkerConfig` utility instead (so that if the unstable function changes in the future we'll only have a single place to change, isolating the risks in our utility and making the rest of our code more resilient).

playground/multi-worker/__tests__/with-wrangler-warning/multi-worker.spec.ts
dario-piotrowicz: again I replied too quickly 😅 

Don't you think showing these warnings on `build` is correct?

I agree with you on `preview` since it relies on the generated worker config file, but on `build` we're still dealing with the user's one

dario-piotrowicz: so I think that line 13 should change:
```diff
-[expectedWarning, expectedWarning],
+[expectedWarning],
```

Also this made me realize that I need to also remove the non-applicable configs from the config file we generate in this PR

dario-piotrowicz:
> Thanks Dario. I've added a few comments that I think need resolving.
> 
> I think this would benefit from some refactoring, renaming etc. and possibly using an approach more similar to the diagnostics in Wrangler. It's not a big deal now though and we can look at those in a follow up PR later.

Thanks 👍 , I am not too familiar with diagnostics in Wrangler, I'll have a look 👍

dario-piotrowicz:
> Thanks 👍 , I am not too familiar with diagnostics in Wrangler, I'll have a look 👍


I think you're referring to this: https://github.com/cloudflare/workers-sdk/blob/8def8c99e1f74f313a3771c827133c59e1b1032b/packages/wrangler/src/config/diagnostics.ts

Any specific reason as to why you think having something more similar to this would be better?

packages/vite-plugin-cloudflare/src/miniflare-options.ts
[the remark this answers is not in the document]
jamesopstad: OK. I thought that you were removing the redundant values from the config and so doing that here would cause problems. I've looked again though and it seems you're not doing that so I see your point. I do think we need to remove the values in dev and build though.

playground/multi-worker/__tests__/with-wrangler-warning/multi-worker.spec.ts
jamesopstad: Yes, you're right. We should warn in build but not preview.

dario-piotrowicz: yeah I made sure that's what's happening now 🙂👍 (https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/112/commits/9a39b9f2459603d856e81211e12a663f2b615dde)

jamesopstad:
> > Thanks 👍 , I am not too familiar with diagnostics in Wrangler, I'll have a look 👍
> 
> I think you're referring to this: https://github.com/cloudflare/workers-sdk/blob/8def8c99e1f74f313a3771c827133c59e1b1032b/packages/wrangler/src/config/diagnostics.ts
> 
> Any specific reason as to why you think having something more similar to this would be better?

I just generally think that if there's already an established pattern for something then it's worth a look. Like I said, it's not a big deal and what you've done is good.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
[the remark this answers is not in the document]
dario-piotrowicz: yes I was considering that, we can totally do it and keep the existing flow/logic intact, `readWorkerConfig` could return both the raw/original config alongside the sanitized/fixed config to use

then the caller can decide which one to use or how (or more likely we stick both on the `resolvedPluginConfig` object and they get picked up and used accordingly in the various places)

would that work for you?

dario-piotrowicz:
> > > Thanks 👍 , I am not too familiar with diagnostics in Wrangler, I'll have a look 👍
> > 
> > 
> > I think you're referring to this: https://github.com/cloudflare/workers-sdk/blob/8def8c99e1f74f313a3771c827133c59e1b1032b/packages/wrangler/src/config/diagnostics.ts
> > Any specific reason as to why you think having something more similar to this would be better?
> 
> I just generally think that if there's already an established pattern for something then it's worth a look. Like I said, it's not a big deal and what you've done is good.

ah ok, that makes sense to me 🙂, the way you put it sounded to me like you were suggesting that there were clear benefits/improvements from that approach (the only one I noticed after a quick glance was the nesting relationship there, which I think doesn't apply here).

after I address the rest of the feedback here I will have a quick play to see how that structure would translate here 👍

packages/vite-plugin-cloudflare/src/miniflare-options.ts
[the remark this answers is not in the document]
jamesopstad: Sure, we can do that if you like. Personally, I don't see any problem with using `unstable_readConfig` directly in preview. It's arguably better because we want preview to reflect what happens when you deploy and that's the function that will be called when you run `wrangler deploy`.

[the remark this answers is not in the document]
dario-piotrowicz: The fact is that we've already seen `unstable_readConfig` break under us, so having our own wrapper around it feels very nice to me

that being said, it doesn't really add a lot of value and I wouldn't be too against using `unstable_readConfig` directly in preview...

dario-piotrowicz:
`Diagnostics` collects [warnings and errors](https://github.com/cloudflare/workers-sdk/blob/2e78812ade7cd7361b023c90afe06221a52b79eb/packages/wrangler/src/config/diagnostics.ts#L8-L9), but I think that we do want to collect more than that, we do also want to keep track of the various non-applicable fields so that we can potentially action upon them later on (if we decide to include ad-hoc handling of some non-applicable fields).

We also don't have the issue of wanting/needing to have a tree-structure, since I think that that is useful/needed when there are various environments but that in our case we only want to warn about the fields that apply to the current environment (we did discuss about this sort of thing; not to overwhelm/bother devs we said that we only want to warn about fields that are actually in use, not everything present in the config file, right?)

So I don't think that trying to copy/follow what `Diagnostics` does would necessarily make our code better and that it would basically boil down to change the code to have a more object oriented code structure, which I would not be opposed to (although I also don't see a huge benefit from that).

So I am again not completely sure if following wrangler's structure is necessary/beneficial here. I am happy to proceed with whatever you prefer, I don't know if you have in mind a specific refactoring (e.g. refactor things not to use `getWorkerConfig` in the `miniflare-options.ts` and have a single unique validation step for build/dev in which we have our own version of `Diagnostics`) if you do, would you maybe like to give your idea a try and open a PR against my branch or even a new PR against main itself so that we can compare and decide which version to go with? (I'm completely fine if we decide to close this PR and go with a different approach).

Or if it's not a huge concern we can keep the implementation as is for now, again I'm happy with whatever 🙂

jamesopstad:
> So I don't think that trying to copy/follow what `Diagnostics` does would necessarily make our code better and that it would basically boil down to change the code to have a more object oriented code structure, which I would not be opposed to (although I also don't see a huge benefit from that).

Don't worry about copying `Diagnostics`. You're right, it's not the best approach for this.

packages/vite-plugin-cloudflare/src/__tests__/get-warning-for-workers-resolved-configs.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+					nonApplicable: {
+						replacedByVite: new Set([]),
+						notRelevant: new Set(['site']),
+						overridden: new Set(),
+					},
+					raw: {} as any,
+				},
+			],
+		});
+		if (process.platform === 'win32') {
```
petebacondarwin, line 151: Generally what we do with windows and snapshots is to post process the warning text to convert any windows paths to unix ones. But I guess this is also OK, although it means that to update the snapshots you must either run locally on Windows with `-u` (which is a pan) or manually change the snapshot (which is not recommended).

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -68,6 +69,11 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin {
 							}
 						: undefined,
 				builder: {
+					// Note: when running `vite build` our plugin is called once per environment, this is usually not
+					//       a problem, but if we present logs etc those would get duplicated, so in order to avoid such
+					//       duplication we set this flag, we can come up with some different solution for the duplicated
+					//       logs if this config becomes problematic for whatever reason
+					sharedConfigBuild: true,
```
petebacondarwin, line 76: How about we only output warnings for the environment being built?
Then we wouldn't get the duplication and this wouldn't be needed.

I am concerned what effect this setting could have on other plugins or user config that expects to be run for each environment.

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -19,25 +19,15 @@ export interface PluginConfig extends Partial<PluginWorkerConfig> {
...
+export type AssetsOnlyConfig = WorkersConfigurations.SanitizedWorkerConfig & {
+	assets: Defined<WorkersConfigurations.SanitizedWorkerConfig['assets']>;
 };
 
-export type WorkerConfig = Unstable_Config & {
-	name: Defined<Unstable_Config['name']>;
-	main: Defined<Unstable_Config['main']>;
+export type WorkerConfig = WorkersConfigurations.SanitizedWorkerConfig & {
+	name: Defined<WorkersConfigurations.SanitizedWorkerConfig['name']>;
+	main: Defined<WorkersConfigurations.SanitizedWorkerConfig['main']>;
```
petebacondarwin, line 28: NIT: perhaps move these to `workers-configs.ts`?

packages/vite-plugin-cloudflare/src/workers-configs.ts
```diff
@@ -0,0 +1,353 @@
...
+import * as path from 'node:path';
+import { unstable_readConfig } from 'wrangler';
+import { name } from '../package.json';
+import type { AssetsOnlyConfig, WorkerConfig } from './plugin-config';
+import type { Optional } from './utils';
+import type { Unstable_Config } from 'wrangler';
+
+type RawWorkerConfig = Unstable_Config;
+
+export namespace WorkersConfigurations {
```
petebacondarwin, line 12: Isn't `namespace` sort of deprecated in TS now?

packages/vite-plugin-cloudflare/src/__tests__/get-warning-for-workers-resolved-configs.spec.ts
dario-piotrowicz: ah ok, yours sounds like a better approach! I'll try that 🙂👍 

(yeah here I've been manually updating the snapshot 😓)

packages/vite-plugin-cloudflare/src/workers-configs.ts
dario-piotrowicz: oof... that's embarrassing 🙈  yeah they seem to be, somehow that completely flew under my radar (I haven't used them a lot)

they seemed like such a nice way to scope similar concepts here and avoid naming collisions 😓 

I'll have to go back to the drawing board and use plain unspaced types 👍

petebacondarwin: Although there are other people who disagree... https://www.reddit.com/r/typescript/comments/1d3zpow/comment/l6bn086/

packages/vite-plugin-cloudflare/src/plugin-config.ts
dario-piotrowicz: I think the line here is quite blurry, because these `WorkerConfig`s are relate/used by to the plugin itself 🤔

packages/vite-plugin-cloudflare/src/workers-configs.ts
```diff
@@ -0,0 +1,353 @@
...
+	],
+	/**
+	 * All the configs that get overridden by our plugin
+	 */
+	overridden: ['rules'],
+} as const;
+
+/**
+ * The non applicable configs that can be and default to `undefined`
+ */
```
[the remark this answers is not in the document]
dario-piotrowicz: `can be` means that those fields can be `undefined`, as not all fields can be `undefined`, for example even if not specified by the user `build` is always defined in an `Unstable_Config` object.

mh... but being able to be `undefined` and defaulting to `undefined` are probably practically the same thing... so maybe it's ok to just simplify the comment 👍

packages/vite-plugin-cloudflare/src/workers-configs.ts
```diff
@@ -0,0 +1,353 @@
...
+	opts?.visitedConfigPaths?.add(configPath);
+
+	if (opts?.isEntryWorker && !config.main) {
+		assert(
+			config.assets,
+			`No main or assets field provided in ${config.configPath}`,
+		);
+
+		return {
+			raw: config,
```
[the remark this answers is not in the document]
dario-piotrowicz: yeah this is not right, not sure why, probably some mistake from refactoring, thanks for noticing it! 👍

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -134,13 +75,23 @@ export function resolvePluginConfig(
...
-		return { ...entryConfigResult, configPaths, persistState };
+	if (entryWorkerResolvedConfig.type === 'assets-only') {
+		return {
+			...entryWorkerResolvedConfig,
+			configPaths,
+			persistState,
+			rawConfigs: {
+				entryWorker: entryWorkerResolvedConfig,
+			},
+		};
```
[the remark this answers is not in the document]
dario-piotrowicz: good spot! 👍

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -1,8 +1,8 @@
 import assert from 'node:assert';
-import * as fs from 'node:fs';
 import * as path from 'node:path';
 import * as vite from 'vite';
-import { unstable_readConfig } from 'wrangler';
+import { findWranglerConfig, getWorkerConfig } from './workers-configs';
+import type { WorkersConfigurations } from './workers-configs';
 import type { Unstable_Config } from 'wrangler';
```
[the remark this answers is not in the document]
dario-piotrowicz: this has already been removed for a while, not sure what happened here 😕  (maybe I didn't push something up?)

packages/vite-plugin-cloudflare/src/__tests__/get-warning-for-workers-resolved-configs.spec.ts
dario-piotrowicz: updated 🙂

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -108,6 +114,16 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin {
...
 			resolvedViteConfig = config;
+			const runningPreview =
+				config.command === 'serve' && config.isProduction === true;
+			if (!runningPreview) {
+				const workersConfigsWarning = getWarningForWorkersConfigs(
+					resolvedPluginConfig.rawConfigs,
+				);
+				if (workersConfigsWarning) {
+					console.warn(workersConfigsWarning);
+				}
```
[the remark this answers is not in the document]
dario-piotrowicz: good call, done 🙂

packages/vite-plugin-cloudflare/src/miniflare-options.ts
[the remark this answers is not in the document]
dario-piotrowicz: resolving as `unstable_readConfig` is now being used for preview as you suggested

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: that's a good point, we could do that but one very nice thing here is that we generate a single warning that combines all the non-applicable fields from all the various workers and avoids multiple separate warnings and collates all the info together nicely 😕 

Since you're worried about this Vite option I removed it and made sure that we show the warning only once by using a boolean flag: https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/112/commits/143d13f3b09fbdf4b17c90fe7dd26fb9f87ff999 please have a look and let me know if this feels better for you

@jamesopstad what do you think? are you ok with this?

jamesopstad: The `boolean` flag is fine with me. I think the `sharedConfigBuild` will become the default at some point in the future so we can change things then.

packages/vite-plugin-cloudflare/src/workers-configs.ts
dario-piotrowicz: namespaces removed 🙂  https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/112/commits/e9710a3d9f54ad5c2f149b9103c31c7503651b8d

petebacondarwin:
Still LGTM

PR #113, opened 2024-12-16
Support Workflows

### LINKED ISSUE #65: Support Workflows

jamesopstad:
resolves #65

Few things to note:
- I initially created a more complex implementation that held onto class instances across requests (similar to the Durable Objects implementation). This turned out to not be necessary because the Workflow `create` method is a one time trigger and the instance ID is unique per Workflow (https://developers.cloudflare.com/workflows/build/rules-of-workflows/#instance-ids-are-unique). I therefore simplified the implementation.
- Changes in the user code won't stop an existing Workflow instance from continuing to execute. This behaviour differs from Wrangler, which restarts Miniflare on any code changes. I have asked the Workflows team for clarification on what happens in production when a new version of the Workflow is deployed while an existing Workflow instance is running.

### DIFF
--- a/packages/vite-plugin-cloudflare/src/miniflare-options.ts
+++ b/packages/vite-plugin-cloudflare/src/miniflare-options.ts
@@ -128,22 +128,26 @@ function getWorkerToDurableObjectClassNamesMap(
 	return workerToDurableObjectClassNamesMap;
 }
 
-function getWorkerToWorkflowClassNamesMap(
+function getWorkerToWorkflowEntrypointClassNamesMap(
 	workers: Array<Pick<WorkerOptions, 'workflows'> & { name: string }>,
 ) {
-	const workerToWorkflowClassNamesMap = new Map(
+	const workerToWorkflowEntrypointClassNamesMap = new Map(
 		workers.map((workerOptions) => [workerOptions.name, new Set<string>()]),
 	);
 
 	for (const worker of workers) {
 		for (const value of Object.values(worker.workflows ?? {})) {
 			if (value.scriptName) {
-				const classNames = workerToWorkflowClassNamesMap.get(value.scriptName);
+				const classNames = workerToWorkflowEntrypointClassNamesMap.get(
+					value.scriptName,
+				);
 				assert(classNames, missingWorkerErrorMessage(value.scriptName));
 
 				classNames.add(value.className);
 			} else {
-				const classNames = workerToWorkflowClassNamesMap.get(worker.name);
+				const classNames = workerToWorkflowEntrypointClassNamesMap.get(
+					worker.name,
+				);
 				assert(classNames, missingWorkerErrorMessage(worker.name));
 
 				classNames.add(value.className);
@@ -151,7 +155,7 @@ function getWorkerToWorkflowClassNamesMap(
 		}
 	}
 
-	return workerToWorkflowClassNamesMap;
+	return workerToWorkflowEntrypointClassNamesMap;
 }
 
 // We want module names to be their absolute path without the leading slash
@@ -360,8 +364,8 @@ export function getDevMiniflareOptions(
 		getWorkerToWorkerEntrypointNamesMap(userWorkers);
 	const workerToDurableObjectClassNamesMap =
 		getWorkerToDurableObjectClassNamesMap(userWorkers);
-	const workerToWorkflowClassNamesMap =
-		getWorkerToWorkflowClassNamesMap(userWorkers);
+	const workerToWorkflowEntrypointClassNamesMap =
+		getWorkerToWorkflowEntrypointClassNamesMap(userWorkers);
 
 	const logger = new ViteMiniflareLogger(viteConfig);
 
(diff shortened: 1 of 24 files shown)


playground/workflows/src/index.ts
```diff
@@ -0,0 +1,60 @@
...
+				instance = await env.MY_WORKFLOW.create();
+			} else {
+				instance = await env.MY_WORKFLOW.create({ id });
+			}
+
+			return Response.json({
+				id: instance.id,
+				status: await instance.status(),
+			});
+		} else if (url.pathname === '/get') {
```
dario-piotrowicz, line 44: Nit: given the early returns I think that there's room to simplify this function by removing the various `else` statements:
Screenshot 2024-12-18 at 17 02 59

I find the version without the multiple `else`s easier to read, but this is really minor and subjective, so if you prefer your version I'm 1000% cool with that, I just figured I could offer an alternative 🙂

playground/workflows/__tests__/workflows.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+	});
+
+	await vi.waitFor(
+		async () => {
+			expect(await getJsonResponse(`/get?id=${instanceId}`)).toEqual({
+				status: 'running',
+				output: [{ output: 'First step result' }],
+			});
+		},
+		{ timeout: 5000 },
```
dario-piotrowicz, line 22: Here and below, can you explain why such a long timeout is needed?

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -126,6 +128,36 @@ function getWorkerToDurableObjectClassNamesMap(
...
+function getWorkerToWorkflowEntrypointClassNamesMap(
+	workers: Array<Pick<WorkerOptions, 'workflows'> & { name: string }>,
+) {
+	const workerToWorkflowEntrypointClassNamesMap = new Map(
+		workers.map((workerOptions) => [workerOptions.name, new Set<string>()]),
+	);
+
+	for (const worker of workers) {
+		for (const value of Object.values(worker.workflows ?? {})) {
+			if (value.scriptName) {
```
dario-piotrowicz, line 140: nit: you're not testing this if branch, it would be great to do so, do you think it'd add too much complexity to the playground app?

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
```diff
@@ -33,6 +46,8 @@ const DURABLE_OBJECT_KEYS = [
 	'webSocketError',
 ] as const;
 
+const WORKFLOW_ENTRYPOINT_KEYS = ['run'] as const;
```
dario-piotrowicz, line 49:
```ts
const WORKFLOW_ENTRYPOINT_KEYS = ['run'] as const satisfies (keyof WorkflowEntrypoint)[];
```

overkill? 🤔

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
```diff
@@ -329,3 +344,39 @@ export function createDurableObjectWrapper(
...
+
+	for (const key of WORKFLOW_ENTRYPOINT_KEYS) {
+		Wrapper.prototype[key] = async function (...args: unknown[]) {
+			const entryPath = this.env.__VITE_ENTRY_PATH__;
+			const ctor = (await getWorkerEntryExport(
+				entryPath,
+				className,
+			)) as WorkflowEntrypointConstructor;
+			const userEnv = stripInternalEnv(this.env);
+			const instance = new ctor(this.ctx, userEnv);
```
dario-piotrowicz, line 361: sorry I see that we do this for entrypoints and durable objects, I must have missed this before
(and sorry as I feel like this might be a silly question 😓)

could you explain why we create a new instance of the entry export for every key and call? is this correct? don't we end up with a bunch of different instances of ctor?

dario-piotrowicz (approved):
Awesome stuff 😄

playground/workflows/src/index.ts
jamesopstad: d5aae06

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
jamesopstad: I just did it like this to make it consistent with the others. I could refactor `createWorkflowEntrypointWrapper` to use the `run` method directly if you prefer but then we'll have to refactor again if they add additional methods.

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
jamesopstad: See the comment in the description:

_I initially created a more complex implementation that held onto class instances across requests (similar to the Durable Objects implementation). This turned out to not be necessary because the Workflow create method is a one time trigger and the instance ID is unique per Workflow (https://developers.cloudflare.com/workflows/build/rules-of-workflows/#instance-ids-are-unique). I therefore simplified the implementation._

From my testing, it seems that calling `workflow.create` creates the instance and then, from that point on, the same ID will always refer to the same instance. So this code actually only gets called once per ID.

For Durable Objects, we do hold on to the same instance unless the user code changes. For `WorkerEntrypoint`, I think it is appropriate to create a new instance on each call as they are stateless.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: We can create a second playground like we do for Durable Objects to test cross-worker Workflows.

playground/workflows/__tests__/workflows.spec.ts
jamesopstad: I based this off some tests in `workers-sdk` so used the timeouts from there. We can shorten it if you like.

dario-piotrowicz: I see, I was just wondering, I thought there might have been a specific reason for this (e.g. you saw it failing in CI or something)

I'm fine with it 👍

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
dario-piotrowicz: Ah yeah sorry I didn't notice we had the other `_KEYS` like this too

This is really not a big deal and I'm totally happy to keep thing consistent 👍

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
dario-piotrowicz: Ah I see, yeah I missed the `kInstance` caching/check for DurableObjects

Ok makes sense 🤔 

Thanks for the explanation 😄

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: b5ec843

PR #115, opened 2024-12-16
Only use router worker when assets is enabled

### LINKED ISSUE #114: Only use router worker when `assets` is enabled

jamesopstad:
resolves #114

This change ensures that the user has to explicitly enable `assets` in their config in order to get the assets functionality.


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -209,21 +209,19 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin {
...
 				}) as any;
 			});
 
 			return () => {
 				viteDevServer.middlewares.use((req, res, next) => {
-					if (!middleware) {
-						next();
-						return;
-					}
-
```
jamesopstad, line 226: This isn't directly related to this PR but was removed as it's unnecessary.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -241,11 +239,6 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin {
 
 			return () => {
 				vitePreviewServer.middlewares.use((req, res, next) => {
-					if (!middleware) {
-						next();
-						return;
-					}
-
```
jamesopstad, line 248: This isn't directly related to this PR but was removed as it's unnecessary.

packages/vite-plugin-cloudflare/src/constants.ts
```diff
@@ -1,9 +1,3 @@
-import type { Miniflare } from 'miniflare';
-
 export const ROUTER_WORKER_NAME = '__router-worker__';
 export const ASSET_WORKER_NAME = '__asset-worker__';
 export const ASSET_WORKERS_COMPATIBILITY_DATE = '2024-10-04';
```
dario-piotrowicz, line 3: sorry it's unrelated to this PR, but `ASSET_WORKER_NAME` and `ASSET_WORKERS_COMPATIBILITY_DATE` are only used in the `miniflare-options.ts` file, maybe they could be simply moved there?

packages/vite-plugin-cloudflare/src/dev.ts
```diff
@@ -0,0 +1,22 @@
...
+import type { ResolvedPluginConfig } from './plugin-config';
+import type { Miniflare } from 'miniflare';
+
+export function getDevEntryWorker(
+	resolvedPluginConfig: ResolvedPluginConfig,
+	miniflare: Miniflare,
+) {
+	const entryWorkerConfig =
+		resolvedPluginConfig.type === 'assets-only'
+			? resolvedPluginConfig.config
```
dario-piotrowicz, line 12: nit: maybe we could add a comment mentioning why we pick the `resolvedPluginConfig.config` if the type is `assets-only`?

dario-piotrowicz (approved):
looks good to me 🙂

PR #120, opened 2024-12-19
Revise output directories

jamesopstad:
Think I've finally got the right approach for output directories. The behaviour is now as follows:

- If there is only a `client` environment then the output directory will resolve in the following order of priority: `environments.client.build.outDir`, `build.outDir`, `'dist'`
- If there are Worker environments (and possibly also a `client` environment) then the output directories will resolve in the following order of priority: `environments[environmentName].build.outDir`, `path.join(build.outDir, environmentName)`, `path.join('dist', environmentName)`

The `assets.directory` path in the output `wrangler.json` is now also correctly derived as the relative path between the Worker and `client` output directories.


playground/spa-with-api/.gitignore
```diff
@@ -0,0 +1,2 @@
+custom-root-output-directory
+custom-client-output-directory
```
dario-piotrowicz, line 2:
```suggestion
custom-client-output-directory

```

playground/multi-worker/.gitignore
```diff
@@ -1 +1,2 @@
-custom-output-directory
\ No newline at end of file
+custom-root-output-directory
+custom-environment-output-directory
```
dario-piotrowicz, line 2:
```suggestion
custom-environment-output-directory

```

packages/vite-plugin-cloudflare/src/utils.ts
```diff
@@ -1,4 +1,18 @@
+import * as path from 'node:path';
 import { Request as MiniflareRequest } from 'miniflare';
+import * as vite from 'vite';
+
+export function getOutputDirectory(
```
dario-piotrowicz, line 5: I think that adding unit tests for this utilities would be very very simple and very very useful (both as tests but especially as documentation, practically converting your PR description into tests)

dario-piotrowicz (approved):
All good to me

as I mentioned I'd really love the newly introduced function to be unit tested though 😄 
(and if you did I'd be fine in dropping one of the playground new variations not to go overboard, but we can also keep them as they don't really hurt)

packages/vite-plugin-cloudflare/src/utils.ts
jamesopstad: [7ec5592](https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/120/commits/7ec55923793372b63b64e49d14e113b46655eb5d)

PR #123, opened 2024-12-24
Re-enable Node.js compatibility

petebacondarwin:
It turned out that the problems were arising because we were also creating virtual modules for Node.js aliases that were already actually expected to be external.

Fixes https://github.com/flarelabs-net/vite-plugin-cloudflare/issues/63

(I still think there is a potential bug in Vite, where you cannot have a virtual module that resolves to an external import, but that become moot by avoiding such cases here.)

### DIFF
--- a/packages/vite-plugin-cloudflare/src/node-js-compat.ts
+++ b/packages/vite-plugin-cloudflare/src/node-js-compat.ts
@@ -74,12 +74,14 @@ export function injectGlobalCode(
  * So instead we alias these to a virtual module, which are then handled with environment specific code in the `resolveId` handler
  */
 export function getNodeCompatAliases() {
-	return Object.fromEntries(
-		Object.keys(preset.alias).map((from) => [
-			from,
-			CLOUDFLARE_VIRTUAL_PREFIX + from,
-		]),
-	);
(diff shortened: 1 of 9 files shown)


packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -97,12 +99,10 @@ export function resolveNodeAliases(source: string, workerConfig: WorkerConfig) {
 	const from = source.slice(CLOUDFLARE_VIRTUAL_PREFIX.length);
 	const alias = preset.alias[from];
 
-	if (alias) {
-		return {
-			id: alias,
-			external: preset.external.includes(alias),
-		};
+	if (alias && preset.external.includes(alias)) {
+		throw new Error(`Alias to external: ${source} -> ${alias}`);
```
petebacondarwin, line 103: I don't know if this is really necessary.
I haven't seen this error been thrown yet.
We can remove it if it is valid use case.

dario-piotrowicz (approved):
Looks good to me 🙂

jamesopstad (approved):
Great! Glad there was a straightforward solution.

PR #125, opened 2025-01-03
feat: run writeDeployConfig in buildEnd

jacob-ebey:
This allows for you to have a custom `builder.buildApp` and still get the necessary files written to disk to support the `preview` command.


jamesopstad:
Thanks Jacob. I agree we should move it out of `builder.buildApp` but I don't think `buildEnd` is the right place as it runs before the output files are written. I originally had it in `closeBundle` but found that this was running once for each environment. I'll check with the Vite team if this is expected.

PR #127, opened 2025-01-14
Support environments in Worker configs

### LINKED ISSUE #29: Respect environments when reading Wrangler config files
The configuration loaded from Wrangler config files can be environment dependent (https://developers.cloudflare.com/workers/wrangler/environments/). We need to decide how the environment should be determined and what the fallback behaviour should be. Possible options are:

1. Reading `viteConfig.mode` (https://main.vitejs.dev/config/shared-options.html#mode)
2. An additional environment option in the Vite config (this could be global or for each worker)

Utilising the mode that Vite is running in makes most sense to me as you get `development` and `production` by default and can run in additional modes with a config or command line option. This provides the least friction for the most common use case of binding to different resources in `development` and `production` as well as additional flexibility. We would also need to fallback somehow if the specified environment is not specified in the Wrangler config.

It would be good to understand more about how this feature is used and how we can best support it.

jamesopstad:
resolves #29

This adds support for the `CLOUDFLARE_ENV` environment variable to determine the environment that is used in the Worker config.

When updating the versions of Wrangler and Miniflare, I had to make changes to the Workflow tests to reflect changes in the implementation. I have also updated all the compatibility dates to the most recent.

Note that the environment variable will not work if provided in a `.env` file. Vite only automatically provides values from `.env` (prefixed with `VITE_`) to code rather than config. The `loadEnv` function (https://vite.dev/guide/api-javascript.html#loadenv) needs to be used to access variables from `.env` elsewhere (by default this only loads env variables prefixed with `VITE_`).

Do you think we should call `loadEnv` in the plugin in so that the value can be read from `.env` files?


playground/cloudflare-env/__tests__/custom-env/cloudflare-env.spec.ts
```diff
@@ -0,0 +1,6 @@
+import { expect, test } from 'vitest';
+import { getTextResponse } from '../../../__test-utils__';
+
+test('returns the correct custom-env var when CLOUDFLARE_ENV=custom-env', async () => {
+	expect(await getTextResponse()).toEqual('Custom env var');
+});
```
jamesopstad, line 6: I tried to use `vi.stubEnv` (https://vitest.dev/api/vi.html#vi-stubenv) here instead of changing `process.env` in the Vite config. It didn't work though (I think because the Vite config has already run by the time we run the test).

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -24,11 +24,16 @@ export interface PluginConfig extends Partial<PluginWorkerConfig> {
 type Defined<T> = Exclude<T, undefined>;
 
 export type AssetsOnlyConfig = SanitizedWorkerConfig & {
+	topLevelName: Defined<SanitizedWorkerConfig['topLevelName']>;
+	name: Defined<SanitizedWorkerConfig['name']>;
+	compatibility_date: Defined<SanitizedWorkerConfig['compatibility_date']>;
```
dario-piotrowicz, line 29: I wonder if `topLevelName`, `name` and `compatibility_date` should be in their own shared base type

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -24,11 +24,16 @@ export interface PluginConfig extends Partial<PluginWorkerConfig> {
...
 	assets: Defined<SanitizedWorkerConfig['assets']>;
 };
 
 export type WorkerConfig = SanitizedWorkerConfig & {
+	topLevelName: Defined<SanitizedWorkerConfig['topLevelName']>;
 	name: Defined<SanitizedWorkerConfig['name']>;
+	compatibility_date: Defined<SanitizedWorkerConfig['compatibility_date']>;
```
dario-piotrowicz, line 36: removing the `compatibility_data` here and in `AssetsOnlyConfig` doesn't seem to break anything, is this really necessary?

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -24,11 +24,16 @@ export interface PluginConfig extends Partial<PluginWorkerConfig> {
...
 export type AssetsOnlyConfig = SanitizedWorkerConfig & {
+	topLevelName: Defined<SanitizedWorkerConfig['topLevelName']>;
+	name: Defined<SanitizedWorkerConfig['name']>;
+	compatibility_date: Defined<SanitizedWorkerConfig['compatibility_date']>;
 	assets: Defined<SanitizedWorkerConfig['assets']>;
 };
 
 export type WorkerConfig = SanitizedWorkerConfig & {
+	topLevelName: Defined<SanitizedWorkerConfig['topLevelName']>;
 	name: Defined<SanitizedWorkerConfig['name']>;
```
dario-piotrowicz, line 35: could we actually not rely on `name` at all anymore and simply always just use `topLevelName`?

(I'm worried that we'll end up sometimes using `name` and sometimes using `topLevelName` potentially confusing things and even potentially making debugging harder)

dario-piotrowicz (approved):
Looks good to me, I just left some minor comments 🙂

dario-piotrowicz:
> Do you think we should call loadEnv in the plugin in so that the value can be read from .env files?

That sounds like a good idea, wrangler does read .env files for config purposes so this would be consistent with that

But I don't know if `loadEnv` picks up more files than wrangler (does it load .env.local files? 🤔) ideally this if were implemented I think it should work as closely as possible to the way wrangler does

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: [793f14b](https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/127/commits/793f14b322e8e534cd201ab7694e4a3946cdd5d5)

I've also updated the types for the plugin config in this file to match those in the docs PR as that won't be merged before we migrate to workers-sdk.

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: I think that as there is now an assertion for the `compatibility_date` it's good for the types to reflect that it will always be defined. Also means that you can't accidentally remove the assertion without a type error.

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: We do need both. `name` is what is passed to Miniflare and used in the `wrangler.json` output whereas `topLevelName` is what's used as the default environment name (and therefore default output subdirectory).

jamesopstad:
> That sounds like a good idea, wrangler does read .env files for config purposes so this would be consistent with that
> 
> But I don't know if `loadEnv` picks up more files than wrangler (I don't think it loads .env.local files for example: [wrangler code](https://github.com/cloudflare/workers-sdk/blob/dceb19608798c2080dc3aa6cfac6f499473f6fb1/packages/wrangler/src/config/index.ts#L222-L232)) ideally if this were implemented I think it should work as closely as possible to the way wrangler does? 🤔

This is only for use within the plugin itself though so we would only use it to get the `CLOUDFLARE_ENV` value. Vite already provides `.env` variables to user code (if they're prefixed with `VITE_`).

packages/vite-plugin-cloudflare/src/plugin-config.ts
dario-piotrowicz: and we want it to be there because it's needed in the file we generate on builds right?

ok makes sense to me, thanks 🙂

packages/vite-plugin-cloudflare/src/plugin-config.ts
dario-piotrowicz: yes, I thought that we could also just pass `topLevelName` to miniflare as well, but that would break service bindings wouldn't it? 😓👍

dario-piotrowicz: (PS: using those two names is a bit unfortunate to me as I think that it feels a bit messy... but I guess that it can't be helped)

dario-piotrowicz:
> This is only for use within the plugin itself though so we would only use it to get the `CLOUDFLARE_ENV` value. Vite already provides `.env` variables to user code (if they're prefixed with `VITE_`).

yeah that's a good point, yes I would be in favour of calling `loadEnv` then 👍

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -11,26 +11,40 @@ import type {
...
 
-export interface PluginConfig extends Partial<PluginWorkerConfig> {
-	auxiliaryWorkers?: PluginWorkerConfig[];
+interface EntryWorkerConfig extends BaseWorkerConfig {
+	configPath?: string;
+}
+
+interface AuxiliaryWorkerConfig extends BaseWorkerConfig {
+	configPath: string;
+}
```
petebacondarwin, line 24: There are a lot of Config types!

Given that we have WorkerConfig stuff below, I think it would be cleaner to call these something else.

E.g.

```suggestion
interface EntryWorkerPluginConfig extends BaseWorkerPluginConfig {
	configPath?: string;
}

interface AuxiliaryWorkerPluginConfig extends BaseWorkerPluginConfig {
	configPath: string;
}
```

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -11,26 +11,40 @@ import type {
...
-export type WorkerConfig = SanitizedWorkerConfig & {
+interface BaseConfig {
+	topLevelName: Defined<SanitizedWorkerConfig['topLevelName']>;
 	name: Defined<SanitizedWorkerConfig['name']>;
-	main: Defined<SanitizedWorkerConfig['main']>;
-};
+	compatibility_date: Defined<SanitizedWorkerConfig['compatibility_date']>;
+}
+
+export type AssetsOnlyConfig = SanitizedWorkerConfig &
```
petebacondarwin, line 39: If you only have a single parent then I think you could use inheritance here?

```suggestion
export interface AssetsOnlyConfig extends BaseWorkerConfig {
```

packages/vite-plugin-cloudflare/src/workers-configs.ts
```diff
@@ -312,30 +315,46 @@ export function getWorkerConfig(
 
 	opts?.visitedConfigPaths?.add(configPath);
 
+	assert(
+		config.topLevelName,
+		`No top-level 'name' field provided in ${config.configPath}`,
+	);
+	assert(config.name, `No 'name' field provided in ${config.configPath}`);
```
petebacondarwin, line 322: Perhaps something like this to help?

```suggestion
	assert(config.name, `No 'name' field provided in ${config.configPath}` + env !== undefined ? ` for ${env} environment.` : '.');
```

playground/cloudflare-env/vite.config.custom-env.ts
```diff
@@ -0,0 +1,10 @@
+import { cloudflare } from '@flarelabs-net/vite-plugin-cloudflare';
+import { defineConfig } from 'vite';
+
+export default defineConfig(() => {
+	process.env = { ...process.env, CLOUDFLARE_ENV: 'custom-env' };
```
petebacondarwin, line 5: Does this need to be undone after the call to `cloudflare()` (perhaps in a finally block) to prevent it from leaking into other tests?

petebacondarwin (approved):
Approved with some nits.

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: [595081b](https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/127/commits/595081b39139b5e0c20fc3579c1d7110a89de655)

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -11,26 +11,40 @@ import type {
...
 }
 
 type Defined<T> = Exclude<T, undefined>;
 
-export type AssetsOnlyConfig = SanitizedWorkerConfig & {
-	assets: Defined<SanitizedWorkerConfig['assets']>;
-};
-
-export type WorkerConfig = SanitizedWorkerConfig & {
+interface BaseConfig {
```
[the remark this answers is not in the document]
jamesopstad: [595081b](https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/127/commits/595081b39139b5e0c20fc3579c1d7110a89de655)

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: I agree that the types and there names need a cleanup but I think this needs to happen across both the `plugin-config` and `workers-configs` files. Is it OK if we don't do this in this PR?

I would rather keep the names of the types at the top of this file (`PluginConfig` and above) the same as they are the only public facing types and are referenced in the docs.

packages/vite-plugin-cloudflare/src/workers-configs.ts
jamesopstad: Should I add this for all the config assertions? Could it be misinterpreted as meaning that a `name` has to be provided for the specific environment?

playground/cloudflare-env/vite.config.custom-env.ts
jamesopstad: Hmm. I tried restoring with `finally` but that doesn't work (presumably because it is restored before the variable is read). It's annoying we can't use `vi.stubEnv` (https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/127#discussion_r1914825858) as that would be a lot nicer.

jamesopstad: If you end up using `package.json` scripts for running the tests then we could just pass the environment variable in there.

jamesopstad: I think you're right about this potentially leaking. Now that I have the test that reads the `CLOUDFLARE_ENV` variable from a `.env` file I could just remove this variant. What do you think?

packages/vite-plugin-cloudflare/src/workers-configs.ts
petebacondarwin: I was thinking for all the assertions. Makes it clearer for the user.
These values are the inherited/overridden values, so if they don't exist at all then they were not at the top level either.
Actually I wonder if this particular one would ever be hit since we would fail on the topLevelName first.

petebacondarwin (approved):

packages/vite-plugin-cloudflare/src/workers-configs.ts
jamesopstad: [fbe0fb3](https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/127/commits/fbe0fb3ecc9d76b557a459ec30ca5424c8b3d594)

